PR #12, opened 2015-01-06
Update the 'abstract' Section of PEP 458

vladimir-v-diaz:
Modify the abstract to target non-specialists and skip details of security concepts.


ncoghlan:
@r1chardj0n3s Would you prefer to review and merge the PEP updates as they come in yourself? I merged the title change, but realised it may be better for you to do it as a marker of "Yes, I've seen this change".

r1chardj0n3s:
I'm not the only one who needs to see the changes, so I don't think it matters who merges them. I'll do this one tho :)

Issue #22, opened 2015-02-01
PEP 458: Review OpenSSH key rotation scheme for applicability


ncoghlan:
OpenSSH 6.8 includes an interesting new key rotation scheme: http://blog.djm.net.au/2015/02/key-rotation-in-openssh-68.html

I'm not sure there's anything applicable there (I think the N/M key quota model addresses the problem in a different way), but I'm still interested in hearing directly from the TUF team on that.

JustinCappos:
[We have a deadline Tuesday, so I only gave this a quick read now.]

From what I see, I agree that our N/M key quota model addresses the issue
of key rotation.  TUF's scheme also has the advantage that it can remove
outdated / compromised keys whenever clients update.

Anyways, we'll do a deeper read in a few days and will post a response here
if there is something we missed or what to clarify.

We will also add a little more detail to the PEP where it discusses this
topic.

Thanks,
Justin
[on behalf of the TUF team]


---------- Forwarded message ----------

vladimir-v-diaz:
The key continuity scheme outlined in the blog *is* supported in TUF :)

The [relevant section of the tuf-spec](https://github.com/theupdateframework/tuf/blob/develop/docs/tuf-spec.txt#L761-L769) document says:

"To replace a compromised root key or any other top-level role key, the root role signs a new root.json file that lists the updated trusted keys for the role. When replacing root keys, an application will sign the new root.json file with both the new and old root keys until all clients are known to have obtained the new root.json file (a safe assumption is that this will be a very long time or never). There is no risk posed by continuing to sign the root.json file with revoked keys as once clients have updated they no longer trust the revoked key. This is only to ensure outdated clients remain able to update."

Updating `root.json` (which contains all of the top-level keys, including the root role's) is part of the automated update process that clients follow to download the TUF metadata.  Clients might download a new `root.json` if repository maintainers release a new `root.json` to revoke keys or add new ones.  Important to note, is that `root.json` is only updated by clients if they already trust a threshold number of the keys included in the new `root.json`.

The update process is sketched here in words:
https://github.com/theupdateframework/tuf/tree/develop/tuf/client#overview-of-the-update-process

And in code:
https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py#L563-L669

[...]

PR #23, opened 2015-02-02
PEP 458: Add Paragraph to the 'In the Event of a Key Compromise' Subsection.  Addresses Issue #22.

vladimir-v-diaz:
* Add paragraph to discuss how new `root.json` metadata should be signed by PyPI administrators following revocation or replacement of top-level role keys.


lvigdor:
Made minor grammar edits.

PR #30, opened 2015-08-03
Adding the draft status PEP for external dependency expression

tleeuwenburg:
Hoping this will initiating a useful discussion around expressing external depencies within setup.py.

### DIFF
--- a/pep-0497-externals.txt
+++ b/pep-0497-externals.txt
@@ -1,0 +1,252 @@
+PEP: 497
+Title: Enhanced handling of external requirements
+Version: $Revision$
+Last-Modified: $Date$
+Author: Tennessee Leeuwenburg <tleeuwenburg@gmail.com>,
+
+Status: Draft
+Type: Standards Track
+Python-Version 3.6
+Content-Type: text/x-rst
+Created: 03-Aug-2015
+Post-History: 03-Aug-2015
+
+
+Abstract
+========
+
+The canonical method for installing packages is to use pip, whether that is
+referencing PyPI or a local repository. A limitation of the current install
+ecosystem is the relationship with operating-system level dependencies. This
+can result in compilation failures or execution failures. Package maintainers
+may wish to specify the requirements to avoid this issue, and users may
+appreciate earlier stopping and more informative messages about their system.
+
+The responsibility of pip is to stop early and provide useful information about
+missing and not automatically installable dependecies. Ideally it would provide
+a useful user message about how to obtain the required dependencies. The
+responsibility for actual installation lies with the user at this time.
+
+This PEP specifies an extension to the requirements sections of setup.py
+which will allow authors to express these requirements cleanly. In the
+implementation section, a discussion of how this is used to solved problems
+is given.
+
+Here is an example of the proposed syntax, where both operating system and build
+target specifiers are included.
+
+::
+
+    setup(
+        name='uses_scipy_with_atlas_extension'
+        install_requires='yaml, scipy[atlas]',
+        setup_requires='pyfortran',
+        external={
+          setup=[
+            'bin!fortran:dist_name=='ubuntu',
+            'include!kernel.h',
+            'include!liblas.h',
+          ],
+          install=[
+              'lib!liblas1.so'
+          ],
+       }
+    )
+
+It is believe this syntax provides one obvious way to do things which does not
+multiply the complexity or parts beyond those necessary to achieve the goal.
+
+In this example, the user will be presented with a useful install message and
+a process that exist early, as opposed to relying on the compiler to fail
+and provide an error message that may not be readily understood by most users
+
+The mechanism of referencing failed dependencies such as 'kernel.h' to useful
+user messages regarding the actual operating system packages that are relevant
+is discussed in the implementation section, along with the rationale for
+handling that referencing outside of setup.py.
+
+Rationale
+=========
+
+Different environment context have implications for the building, testing,
+installing and execution of software. There are two sides to this coin. First,
+in some environments, additional python packages are needed. Secondly,
+it is useful to specify, e.g., the apt or yum package names which need to be
+in place prior to continuing. The syntax proposed here allows for a readable
+and consise expression of these requirements.
+
+Setuptools and pip should be modified to support the use case of installing
+the appropriate python packages as required depending on the specification.
+
+They should also support aborting an install with a useful error message if
+required dependecies are not installed in the environment. The responsibility
+for actually installing those required packages lies with the user.
+
+Ideally, and error message will display the package manager command required
+to add the missing environment dependencies.
+
+It is expected that this feature would be commonly used, particularly in
+scientific applications which frequently have complex interactions with
+externally provided libraries. An example would be needing a version of
+numpy and scipy both of which have been compiled to be aware of the ATLAS
+compiled set of linear algebra libraries for performance reasons. This sounds
+esoteric but is, in fact, a routinely encountered situation which drives
+people towards using the anaconda scientific python environment since it
+actually solves these issues.
+
+Syntax
+======
+
+The valid context specifiers are taken from the environment markers library.
+This command says that when the operating system matches 'ubuntu', then add
+the yaml==0.3 requirement to the list.
+
+Implementation
+==============
+
+A common user experience is to attempt the installation of software in a
+basic environment, and for the installation to fail during the compilation
+of a dependency. The user is presented with an error message provided by the
+compiler. Not all users have the ability to understand these messages, and
+a common experience is to spend time searching online for the appopriate
+package which is provided by the vendor.
+
+The base implementation will also only provide information about missing
+headers, libraries or executables. This is a step forward in that the error
+message is provided very early in the install process, rather than potentially
+many minutes in to a large compile job.
+
+A key aspect of the usability of this tool is a mechanism to provide more useful
+information to the user about how to obtain those missing resources.
+The proposal is to provide a layer inside of pip which is capable of providing
+information about appropriate package names. It is surprising that vendors
+do not already provide a good tool for this, given that finding appropriate
+header and library files is a common experience in the general linux ecosystem.
+However, it is also a critical aspect of the user experience, and something
+which package maintainers will need to be able to specify precisely. The
+proposed approach would be for pip to supply the ability to install an
+inference layer, usable by users and vendors alike. Usage would be along
+the lines of
+
+    pip install_vendor_magic darwin
+
+This would transform the user messages relating to missing dependencies from
+
+    Could not find libatlas.so
+
+to
+
+    Could not find libatlas.so. Run 'port install scipy +atlas'
+
+
+Discussion
+==========
+
+This section is incomplete. Feedback appreciated on relevant discussion
+points that should be included.
+
+Specification of files (headers, libs, bins) vs packages
+--------------------------------------------------------
+
+There was a discussion about whether specific dependencies (named headers,
+libraries, binaries etc) should be included, or whether general package names
+(such as may be expected by yum, apt, conda) were appropriate. Due to the
+number of package management tools, plus the changing nature of those, it
+was decided that concentrating on the specific files was more useful than
+using package names. It also provides looser coupling between the specification
+of externals and the package management tools.
+
+Location in setup.py rather than requirements.txt
+-------------------------------------------------
+
+Requirements.txt is scoped to the use for local deployments to specify
+requirements precisely for the relevant local environment. Setup.py is the
+appropriate place to specify the information for the package in general.
+
+In future, there may be some scope for including similar directives within
+requirements.txt, but that is out of scope for this PEP.
+
+
+Addition of vendor inference layer
+----------------------------------
+
+Bearing in mind the discussion about using specific dependencies, the ideal
+user experience is to get a useful message about how to get the package
+installed on their system. This means providing a message which is relevant
+for their package management environment (e.g. yum, apt, conda). Each
+distributor would be expected to provide an inference lookup which can be
+installed into pip which would assist with identifying the packages which
+are associated with the named dependencies.
+
+
+Types of environment and vendor
+-------------------------------
+
+Anaconda is not an operation system. Neither is yum, neither is apt. It is the
+maintainer of the packaging system which is pertinent, and the current operating
+system is an imperfect indicator of the relevant package manager. 
+
+
+Directories, categories and locations of dependency files
+---------------------------------------------------------
+
+The proposed mechanism allows introspection of the standard location for
+packages. On *nix environments, these would be (for example), /usr/bin,
+/usr/local/include etc (see https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
+for a discussion of standard directories). Non-standard dependency locations
+would be supported only via the underlying platform's mechanisms for exposing
+these details, such as PATH, LD_LIBRARY_PATH etc. This PEP does not include
+generalised configuration and specification of dependencies in non-standard
+or customised locations.
+
+
+
+Alternative Syntax Concept
+==========================
+
+One early straw man for the syntax is included here for reference during
+discussions. The chosen approach is based off approaching a small dsl for
+handling the specification, which may also point to future directions that
+could be taken in requirements.txt.
+
+::
+
+    setup(
+        install_requires='yaml',
+        setup_requires='pyfortran',
+        os_requires=[
+            ['fortran', type='bin', os='ubuntu', 'target=build'],
+            ['kernel.h', type='header', os='linux', target='build'],
+            ['libblas.o', type='lib', target='run']
+            ['scipy +atlas', type='packagename', target='run', os='darwin']
+        ]
+    )
+
+
+References
+==========
+
+.. [1] PEP 426, Metadata for Python Software Packages, Coghlan, Holth, Stufft
+   (http://www.python.org/dev/peps/pep-0426)
+
+.. [2] PEP 345, Sample Plaintext PEP Template, Warsaw
+   (http://www.python.org/dev/peps/pep-0009)
+
+.. [3] Pull request 29 on interoperability-peps
+
+
+Copyright
+=========
+
+This document has been placed in the public domain.
+
+
+
+..
+   Local Variables:
+   mode: indented-text
+   indent-tabs-mode: nil
+   sentence-end-double-space: t
+   fill-column: 70
+   coding: utf-8
+   End:


pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+Python-Version 3.6
+Content-Type: text/x-rst
+Created: 03-Aug-2015
+Post-History: 03-Aug-2015
+
+
+Abstract
+========
+
+The canonical method for installing packages is to use pip, whether that is
```
rbtcollins: I think this should say 'Python packages', as I don't think this pep is implying or intending to imply that installing e.g. Python3 itself would be a pip responsibility.

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+referencing PyPI or a local repository. A limitation of the current install
+ecosystem is the relationship with operating-system level dependencies. This
```
rbtcollins: Lets say 'Non-Python' dependencies. There are many distributions of things that are not operating systems:
 - npm
 - conda
 - nix
...

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+can result in compilation failures or execution failures. Package maintainers
+may wish to specify the requirements to avoid this issue, and users may
+appreciate earlier stopping and more informative messages about their system.
```
rbtcollins: Might read more smoothly as 'appreciate earlier and more informative error messages.'

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+
+The responsibility of pip is to stop early and provide useful information about
+missing and not automatically installable dependecies. Ideally it would provide
+a useful user message about how to obtain the required dependencies. The
+responsibility for actual installation lies with the user at this time.
+
+This PEP specifies an extension to the requirements sections of setup.py
```
rbtcollins: Lets be broader here.

This PEP can be thought of as specifying an extension to the descriptions of 'distributions' in PEP-426. I think we need to update that // make this be referenced, and reference PEP-426. https://www.python.org/dev/peps/pep-0426/#name

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+        setup_requires='pyfortran',
+        external={
+          setup=[
+            'bin!fortran:dist_name=='ubuntu',
+            'include!kernel.h',
+            'include!liblas.h',
+          ],
+          install=[
+              'lib!liblas1.so'
+          ],
```
rbtcollins: So I've been thinking about this since we spoke, and I think it needs to look like this:

```
    setup(
        name='uses_scipy_with_atlas_extension'
        install_requires=['yaml', 'scipy[atlas]', 'lib!libblas1.so'],
        setup_requires=[
            'pyfortran',
            'bin!fortran:dist_name=="ubuntu"',
            'include!kernel.h',
            'include!liblas.h',
          ],
    }
    )
```

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+       }
+    )
+
+It is believe this syntax provides one obvious way to do things which does not
```
rbtcollins: I think I'd delete these two lines. We can perhaps add a 'discarded alternatives' section.

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+multiply the complexity or parts beyond those necessary to achieve the goal.
+
+In this example, the user will be presented with a useful install message and
+a process that exist early, as opposed to relying on the compiler to fail
```
rbtcollins: s/exist/exits/

The paragraph doesn't say that this only happens in the case that the dependencies are missing :). I think thats an important thing to note.

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+
+The mechanism of referencing failed dependencies such as 'kernel.h' to useful
+user messages regarding the actual operating system packages that are relevant
+is discussed in the implementation section, along with the rationale for
+handling that referencing outside of setup.py.
+
+Rationale
+=========
+
+Different environment context have implications for the building, testing,
```
rbtcollins: contexts?

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+installing and execution of software. There are two sides to this coin. First,
+in some environments, additional python packages are needed. Secondly,
```
rbtcollins: capital P on Python

pep-0497-externals.txt
```diff
@@ -0,0 +1,252 @@
...
+it is useful to specify, e.g., the apt or yum package names which need to be
+in place prior to continuing. The syntax proposed here allows for a readable
+and consise expression of these requirements.
```
rbtcollins: The domain of the two things - Python package names and APT package names are quite different - and we've settled (for now at least) on artifacts rather than the abstract APT package names. So I think this should say something like

'in some environments additional Python packages are needed. Secondly, in some environments non-Python package resources such as header files, libraries, scripts and so on are needed. The syntax proposed here allows describing both sorts of requirements in a single syntax.

pep-0497-externals.txt
tleeuwenburg: I'm not opposed, but I don't think this is obvious either. The original version puts the external deps into a clearly separated area for the purpose; your suggestion pulls the two together. This may be conflating two things
event though it makes it shorter. Perhaps it is worth having a quick chat to decide which version goes in the abstract, and then including the alternative later in the document for discussion.

pep-0497-externals.txt
tleeuwenburg: done

pep-0497-externals.txt
tleeuwenburg: done

pep-0497-externals.txt
tleeuwenburg: done

pep-0497-externals.txt
tleeuwenburg: Added alternatives later in the document, stuck with the original version for now. About to upload an updated version.

tleeuwenburg: done

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+appreciate earlier and more informative messages about their system.
+
+The responsibility of pip is to stop early and provide useful information about
+missing (and not automatically installable) dependencies. Ideally it would provide
+a useful user message about how to obtain the required dependencies. The
+responsibility for actual installation lies with the user at this time.
+
+This PEP proposes a mechanism for extending the requirements section of setup.py
+which will allow authors to express non-Python requirements cleanly. In the
+implementation section, a discussion of how this may be used to solved problems
```
rbtcollins: s/solved/solve/ I think

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+can result in compilation failures or execution failures. Package maintainers
+may wish to specify the requirements to avoid this issue, and users may
+appreciate earlier and more informative messages about their system.
+
+The responsibility of pip is to stop early and provide useful information about
+missing (and not automatically installable) dependencies. Ideally it would provide
+a useful user message about how to obtain the required dependencies. The
+responsibility for actual installation lies with the user at this time.
+
+This PEP proposes a mechanism for extending the requirements section of setup.py
```
rbtcollins: This is a topic switch from pip to setuptools/distutils. This PEP needs to be clear it covers two things:

1. How users specify external dependencies when using setuptools
1. And how they specify them when using pip (e.g. in requirements.txt files)
2. How external dependencies will be communicated between tools (setuptools -> wheel, wheel -> pip, setuptools -> pip, ...)

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+which will allow authors to express non-Python requirements cleanly. In the
+implementation section, a discussion of how this may be used to solved problems
+is given.
+
+Here is an example of the proposed syntax, where both operating system and build
```
rbtcollins: syntax + 'for setuptools'.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+
+They should also support aborting an install with a useful error message if
+required dependecies are not installed in the environment. The responsibility
+for actually installing those required packages lies with the user.
+
+This would allow automatic installation of environment-specific Python
+dependencies and pave the way to improving how users go about installing
+non-Python prerequisites.
+
+Ideally, and error message will display the package manager command required
```
rbtcollins: s/and error/any error/

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+compiled set of linear algebra libraries for performance reasons. This sounds
+esoteric but is, in fact, a routinely encountered situation which drives
+people towards using the alternative packaging for scientific python
+environments.
+
+Syntax
+======
+
+The valid context specifiers are taken from the environment markers library.
+This command says that when the operating system matches 'ubuntu', then add
```
rbtcollins: Either say 'the example', or give a new example here.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+the yaml==0.3 requirement to the list.
+
+Meta: I am not a syntax wizard and will need help from others to properly specify
+the syntax as a formal grammar if that is required.
```
rbtcollins: I'd delete this meta comment.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+
+Essentially, the "!" separator is used to flag key-value pairs which will allow
```
rbtcollins: "The proposed syntax is an external object type, a "!", and then an external object name. The object type can be one of:

- include
- bin
- lib

The object name should be the name of the required thing - search paths are presumed to be used to find the thing.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+
+For example, consider the line
+
+    setup_requires=['bin!fortran:dist_name=="ubuntu"']
+
+The 'bin' specifies one of "bin", "lib" or "header", representing the only
+non-Python dependency types allows. The absence of this prefix implies the the
+package name is a discoverable Python package.
+
+Following the ':' is a comma-separated list of key-operator-value pairs. These
```
rbtcollins: The : is being used to introduce PEP-496 environment markers- rather than a loose description of them here, lets just refer to that PEP.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+numpy and scipy, both of which must have been compiled to be aware of the ATLAS
+compiled set of linear algebra libraries for performance reasons. This sounds
+esoteric but is, in fact, a routinely encountered situation which drives
+people towards using the alternative packaging for scientific python
+environments.
+
+Syntax
+======
+
+The valid context specifiers are taken from the environment markers library.
```
ncoghlan: this can refer to PEP 496 now

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+
+Following the ':' is a comma-separated list of key-operator-value pairs. These
+represent the sufficient trigger conditions to apply the rule. At this stage
+only a very limited set of allowable operators and logics are permitted. The
+goal is to provide just enough support to allow the specification of key
+dependencies for specific environments, not to become a make/build system.
+
+Implementation Concerns
+=======================
+
```
rbtcollins: Lets break this into three subsections: setuptools, wheel, pip

rbtcollins: The setuptools one needs to cover:

setup.py UI

egg_info metadata output (just put it inline into the existing files and fields)

binary dependency UI: will the numpy authors manually write:
'lib!libblas1.so'
or will they hint to setuptools that it it should figure out what was used by the binary
or should setuptools use ldd etc to introspect the result of building an extension and automatically add it?
or perhaps automatically derive it but only add it when users opt-in ?

I don't have a final answer here, and think we'll need some detailed discussion with @jaracao about it.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+Essentially, the "!" separator is used to flag key-value pairs which will allow
+two things: the specification of dependency type and the environment context
+where the rule should apply.
+
+For example, consider the line
+
+    setup_requires=['bin!fortran:dist_name=="ubuntu"']
+
+The 'bin' specifies one of "bin", "lib" or "header", representing the only
```
ncoghlan: s/allows/allowed/

s/the the/the/

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+If the user does not have the required dependencies, the user will be presented
+with a useful install message and a process that exits early, as opposed to
+relying on the compiler to fail and provide an error message that may not be
+readily understood by most users.
+
+The mechanism of referencing failed dependencies such as 'kernel.h' to useful
+user messages regarding the actual operating system packages that are relevant
+is discussed in the implementation section, along with the rationale for
+handling that referencing outside of setup.py.
+
```
rbtcollins: We need to add in here a scoping thing.

This PEP does not (and shouldn't - its a bigger and different problem) address the problem of choosing a particular wheel. It *does* cover the problem of detecting whether a given wheel that you've already chosen is able to be installed and run.

So something like:
"In summary, this PEP covers:
 - specifying external build and run time dependencies for source trees, sdists and (run time only) wheels
 - metadata for describing external build and runtime dependencies

This PEP does not cover:
 - identifying which wheel is likely to be suitable to install on a given Linux platform. That problem, while related, is separate.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+(such as may be expected by yum, apt, conda) were appropriate. Due to the
+number of package management tools, plus the changing nature of those, it
+was decided that concentrating on the specific files was more useful than
+using package names. It also provides looser coupling between the specification
+of externals and the package management tools.
+
+Location in setup.py rather than requirements.txt
+-------------------------------------------------
+
+Requirements.txt is scoped to the use for local deployments to specify
```
ncoghlan: @rbtcollins pointed out that this PEP should be more explicitly targeted at improving the "sdist -> wheel -> install" user experience. That changes the answer a bit here to include pointing out that requirements.txt isn't involved when installing an sdist directly.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+Some capacity to support description of external dependencies (among other
+things) was described in PEPs 314 and 345 (Metadata 1.1 and 1.2). These
+describe directives whichs can be placed in the PKG-INFO file. An implementation
+responding to these directives is pending.
+
+This PEP places the directives in setup.py. This approach is closer to the
+typical approach of package maintainers to expressing requirements and places
+the external dependency specification nearer to the Python dependency
+specification. This PEP would supercede PEPs 314 and 345 with respect to
+external dependencies.
```
ncoghlan: This answer isn't quite right. The problem with PKG-INFO (and metadata 1.x in general) is that it doesn't clearly distinguish "build time metadata passed from developers to their chosen build tool" from "install time metadata passed from a developer's chosen build tool to an end user's chosen install tool".

There's actually an open question for this PEP as to whether the design should be "developer specifies expected external dependencies in the input metadata, build tool may complain if the actual external dependencies go beyond the expected dependencies", or "build tool automatically adds external dependencies to the generated metadata".

I think the PEP gets it right (where external dependencies are only added deliberately, and we make the input metadata richer so we can compare intent and actual dependencies), but the rationale in this answer could be improved by considering the trade-off more explicitly.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+for their package management environment (e.g. yum, apt, conda). Each
+distributor would be expected to provide an inference lookup which can be
+installed into pip which would assist with identifying the packages which
+are associated with the named dependencies.
+
+
+Types of environment and vendor
+-------------------------------
+
+Anaconda is not an operation system. Neither is yum, neither is apt. It is the
```
ncoghlan: s/operation/operating/

pep-0497-externals.txt
rbtcollins: The pip section needs to cover:
 
- use in requirements.txt / constraints.txt files / pip command line
- the fact this doesn't alter (at least as we have specified so far) the wheel platform that looks for on PyPI
- selection / detection of the translation layer module
- when it should call into the translation module

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+---------------------------------------------------------
+
+The proposed mechanism allows introspection of the standard location for
+packages. On *nix environments, these would be (for example), /usr/bin,
+/usr/local/include etc (see https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
+for a discussion of standard directories). Non-standard dependency locations
+would be supported only via the underlying platform's mechanisms for exposing
+these details, such as PATH, LD_LIBRARY_PATH etc. This PEP does not include
+generalised configuration and specification of dependencies in non-standard
+or customised locations.
```
ncoghlan: Perhaps mention the translation layer again here.

pep-0497-externals.txt
rbtcollins: platform abstraction needs to cover:

- how pip/setuptools should access it
- the API for using it
- who is expected to implement it
- how it gets into virtualenvironments


strawman:
import vendorplatform;
p = vendorplatform.Platform()

Vendors (distros, conda, etc) will include a vendorplatform package or module in their base Python. It must be selfcontained (e.g. it can be a .so or a .py or a package, but it can't call into other non-stdlib packages).

virtualenv should be modified to copy that into virtualenvs that it creates.

If vendorplatform can't be imported, the calling code should just not attempt to translate calls.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+        name='uses_scipy_with_atlas_extension'
+        install_requires=['yaml', 'scipy[atlas]', 'lib!libblas1.so'],
+        setup_requires=[
+            'pyfortran',
+            'bin!fortran:dist_name=="ubuntu"',
+            'include!kernel.h',
+            'include!liblas.h',
+          ],
+    )
+
```
ncoghlan: Another advantage to this approach it that any tool which treats setup_requires (et al) as a list of opaque strings should "just work", while a separate field causes problems.

Notably, this change should mostly be forwards compatible with the overall structure of PEP 426.

pep-0497-externals.txt
rbtcollins: wheel section needs to cover:

picking up the runtime deps during the build process
exposing them to pip / other installers

but - I think this is entirely pass-through and nearly a no-op

pep-0497-externals.txt
tleeuwenburg: done

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+package which is provided by the vendor.
+
+The base implementation will also only provide information about missing
+headers, libraries or executables. This is a step forward in that the error
+message is provided very early in the install process, rather than potentially
+many minutes in to a large compile job.
+
+A key aspect of the usability of this tool is a mechanism to provide more useful
+information to the user about how to obtain those missing resources.
+The proposal is to provide a layer inside of pip which is capable of providing
```
rbtcollins: s/inside of/consumed by/

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+
+    pip install vendor_magic darwin
+
+This would transform the user messages relating to missing dependencies from
+
+    Could not find libatlas.so
+
+to
+
+    Could not find libatlas.so. Run 'port install scipy +atlas'
```
rbtcollins: I think this is crossing layers here - I'd expect the error to describe how to install libatlas.so, not how to install scipy which is just one user of it.

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+
+Location in setup.py rather than requirements.txt
+-------------------------------------------------
+
+Requirements.txt is scoped to the use for local deployments to specify
+requirements precisely for the relevant local environment. Setup.py is the
+appropriate place to specify the information for the package in general.
+
+In future, there may be some scope for including similar directives within
+requirements.txt, but that is out of scope for this PEP.
```
rbtcollins: I think because we're extending the definition of 'distribution name' that we need to consider requirements.txt in this spec.

pep-0497-externals.txt
rbtcollins: Oh, and we should be clear that this doesn't provide version information, so <= == etc don't make sense here.

pep-0497-externals.txt
rbtcollins: We need to note that it will flow through to PKG-INFO (but the sections I note we need above will include this)

pep-0497-externals.txt
```diff
@@ -0,0 +1,359 @@
...
+Addition of vendor inference layer
+----------------------------------
+
+Bearing in mind the discussion about using specific dependencies, the ideal
+user experience is to get a useful message about how to get the package
+installed on their system. This means providing a message which is relevant
+for their package management environment (e.g. yum, apt, conda). Each
+distributor would be expected to provide an inference lookup which can be
+installed into pip which would assist with identifying the packages which
+are associated with the named dependencies.
```
rbtcollins: Yah, prose for the platform layer - but needs to move up above the FAQ

pep-0497-externals.txt
rbtcollins: I think we can say that pip etc should /honour/ those standard mechanisms but that its per-platform and not specified by this specification.

pep-0497-externals.txt
tleeuwenburg: done

pep-0497-externals.txt
tleeuwenburg: done

pep-0497-externals.txt
tleeuwenburg: added

pep-0497-externals.txt
tleeuwenburg: thanks

njsmith:
Some discussion of this PEP on distutils-sig: http://thread.gmane.org/gmane.comp.python.distutils.devel/23712

pep-0497-externals.txt
```diff
@@ -0,0 +1,460 @@
...
+        install_requires='yaml, scipy[atlas]',
+        setup_requires='pyfortran',
+        external={
+          setup=[
+            'bin!fortran:dist_name=='ubuntu',
+            'include!kernel.h',
+            'include!liblas.h',
+          ],
+          install=[
+              'lib!liblas1.so'
```
leorochael: Specifying `install` dependencies here on `setup.py` will be very useful, specially with regards to specifying external processes that will be needed by the package (e.g. `bin!pdftk`).

However, regarding dependencies into external shared libraries, it would be better if the build tool generated those automatically.

It's a lot easier for, say, `setuptools` or an extension to it to run `ldd build/lxml_***/etree.so` and figure out the library dependencies and put those in the wheel metadata, than forcing the poor package maintainer to do so to put it into `setup.py`.

In particular, different linux distros will have libraries that are source compatible at build time, but result in shared library dependencies with different versions on the python extension.

Forcing this information to be present in `setup.py` would force the package maintainer to build this package in a variety of platforms just to be able to add the proper `'lib!libsomenthing.so.X.Y:dist_name=="Z"'` for every supported platform.

Whereas if the tooling automatically fed this information to the wheel metadata, not only would the package maintainer be relieved of this chore, but the package builder would be able to chose between different source compatible libraries, or even between static and dynamic linking, and the generated wheel would correctly carry the information needed to install it.

ncoghlan: We don't want to add the dependency automatically because the developer
might want to find a way to remove it, statically link it, or figure out a
suitable distro independent way of describing it, rather than including a
distro specific dependency in the wheel metadata.

It definitely makes sense for the build system to detect and report
discrepancies between stated intent and actual build results when it comes
to external dependencies, but they should still be made explicit rather
than left implicit.

qwcode:
just a minor point of order.. this should be in the "unpublished" directory.  it's not really PEP497 yet, and there's actually something else now taking PEP497

pep-0497-externals.txt
tleeuwenburg: Hi Leonardo,

I am coming to this discussion a bit late, because "life". I was involved
in authoring this PEP although rbt now has carriage of it.

We discussed the issues you mentioned in some depth. I am not myself a
packaging guru or compile expert, but I am a maintainer of small packages
and software engineer. I'm a 'user' of pip for the most part. You're not
wrong in identifying those concerns but I think there is a balance of
issues to consider. I will present my viewpoint based on my use cases:

  (i) Firstly, my view is that the ability to store these things in
setup.py will more or less immediately address a very large proportion of
the really common issues and is well worth doing.

  (ii) Engineering the relevant build tools is much harder than allowing
some simple manual configuration

  (iii) Pulling the dependency gaps out of build tools might work for the
maintainer, but you would still need to store the dependency gap
information in some kind of file. In my use case, the install will often
fail 90 seconds into compiling some dependency, and the goal is to provide
the information back to users much faster than that. This isn't about
'fixing dependency resolution in across wheels and external packaging
systems (e.g. yum/conda), but rather about providing the pragmatic
workarounds required to fix the major pain points.

  (iv) A previous accepted PEP did include some capability to include
metadata in one of the metadata files. To my knowledge it was never
implemented in tools. I think it was just too complicated. On the other
hand, I know of a lot of immediate issues a simple extension to setup.py
would allow

In general, the uses I know of are for 'the exception' rather than 'the
rule'. For example, packages like scipy are "mostly fine" to compile, but
fail in some environments.

I would also consider this might not be mutually exclusive with what you
are suggesting. It may be that the larger engineering work you are
suggesting is justified, I am not sure. I would certainly love it if there
were a two-way bridge between pip and the various external packaging
systems so that e.g. a pip install could also submit its dependencies to
yum for a request to install, however we just don't live in that world.

Thanks,
-Tennessee

On 14 August 2015 at 01:44, Leonardo Rochael Almeida <
notifications@github.com> wrote:

> In pep-0497-externals.txt
> <https://github.com/pypa/interoperability-peps/pull/30#discussion_r36989624>
> :
>
> > +determine whether a named package is internal or external.
> > +
> > +::
> > +
> > +    setup(
> > +        name='uses_scipy_with_atlas_extension'
> > +        install_requires='yaml, scipy[atlas]',
> > +        setup_requires='pyfortran',
> > +        external={
> > +          setup=[
> > +            'bin!fortran:dist_name=='ubuntu',
> > +            'include!kernel.h',
> > +            'include!liblas.h',
> > +          ],
> > +          install=[
> > +              'lib!liblas1.so'
>
> Specifying install dependencies here on setup.py will be very useful,
> specially with regards to specifying external processes that will be needed
> by the package (e.g. bin!pdftk).
>
> However, regarding dependencies into external shared libraries, it would
> be better if the build tool generated those automatically.
>
> It's a lot easier for, say, setuptools or an extension to it to run ldd
> build/lxml_***/etree.so and figure out the library dependencies and put
> those in the wheel metadata, than forcing the poor package maintainer to do
> so to put it into setup.py.
>
> In particular, different linux distros will have libraries that are source
> compatible at build time, but result in shared library dependencies with
> different versions on the python extension.
>
> Forcing this information to be present in setup.py would force the
> package maintainer to build this package in a variety of platforms just to
> be able to add the proper 'lib!libsomenthing.so.X.Y:dist_name=="Z"' for
> every supported platform.
>
> Whereas if the tooling automatically fed this information to the wheel
> metadata, not only would the package maintainer be relieved of this chore,
> but the package builder would be able to chose between different source
> compatible libraries, or even between static and dynamic linking, and the
> generated wheel would correctly carry the information needed to install it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pypa/interoperability-peps/pull/30/files#r36989624>.
>



-- 
--------------------------------------------------
Tennessee Leeuwenburg
http://myownhat.blogspot.com/
"Don't believe everything you think"

tleeuwenburg:
Yes, that's true. I was told to go ahead and use that number but history
has not passed us by. I expect someone else will do this update but I can
manage it if people would like me to.

On 14 September 2015 at 08:01, Marcus Smith <notifications@github.com>
wrote:

> just a minor point of order.. this should be in the "unpublished"
> directory. it's not really PEP497 yet, and there's actually something else
> now taking PEP497
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pypa/interoperability-peps/pull/30#issuecomment-139922162>
> .
>



-- 
--------------------------------------------------
Tennessee Leeuwenburg
http://myownhat.blogspot.com/
"Don't believe everything you think"

tleeuwenburg:
I meant has "now" passed us by. Need more coffee :)

On 18 September 2015 at 09:11, Tennessee Leeuwenburg <tleeuwenburg@gmail.com
> wrote:

> Yes, that's true. I was told to go ahead and use that number but history
> has not passed us by. I expect someone else will do this update but I can
> manage it if people would like me to.
>
> On 14 September 2015 at 08:01, Marcus Smith <notifications@github.com>
> wrote:
>
>> just a minor point of order.. this should be in the "unpublished"
>> directory. it's not really PEP497 yet, and there's actually something else
>> now taking PEP497
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/pypa/interoperability-peps/pull/30#issuecomment-139922162>
>> .
>>
>
>
>
> --
> --------------------------------------------------
> Tennessee Leeuwenburg
> http://myownhat.blogspot.com/
> "Don't believe everything you think"
>



-- 
--------------------------------------------------
Tennessee Leeuwenburg
http://myownhat.blogspot.com/
"Don't believe everything you think"

rbtcollins:
I've got it on my TODO.

ppisar:
Besides dependencies on libraries, header files and programs, I'd like to see support for dependencies on pkg-config modules [http://www.freedesktop.org/wiki/Software/pkg-config/] including their versions.

raiderrobert:
Bringing up this idea again, this may be more possible now than ever before. It was suggested over here: https://github.com/pypa/pipenv/issues/1469#issuecomment-367882974

pep-0497-externals.txt
ncoghlan: Revisiting this idea several years later for https://github.com/pypa/pipenv/issues/2082, and I think if we ever pursue this, we're going to need to go with the "separate fields for external dependencies" approach. The problem is that there are now plenty of tools that expect requirements definitions to comply with the PEP 508 dependency specifier format, and "failed to do anything useful with the supplied external requirements definition" is a more graceful degradation to the status quo behaviour than "threw an exception complaining about a non-compliant dependency specifier".

pep-0497-externals.txt
msarahan: In conda-build, we have a detection step that examines the linkage of any binaries.  For dependencies that get detected there, we match them up with package names in conda where those libraries can be obtained.  If those packages are not part of our "run" dependencies, we warn, but conda-build can also be configured to error out.  This has dramatically improved the quality of our packages.

It's a lot harder to think about how to do this in the general case.  I suppose as long as you only compare your manually specified dependency against detected dependencies, that's pretty viable.  Comparing against arbitrary external package managers to ensure that some external package dependency is listed would be a whole lot harder, though.

PR #54, opened 2015-10-28
build system abstraction PEP

rbtcollins:
[Review on Reviewable](https://reviewable.io/reviews/pypa/interoperability-peps/54)

### DIFF
--- a/build-system-abstraction.rst
+++ b/build-system-abstraction.rst
@@ -1,0 +1,475 @@
+:PEP: XX
+:Title: Build system abstraction for pip/conda etc
+:Version: $Revision$
+:Last-Modified: $Date$
+:Author: Robert Collins <rbtcollins@hp.com>,
+         Nathaniel Smith <njs@pobox.com>
+:BDFL-Delegate: Donald Stufft <donald@stufft.io>
+:Discussions-To: distutils-sig <distutils-sig@python.org>
+:Status: Draft
+:Type: Standards Track
+:Content-Type: text/x-rst
+:Created: 26-Oct-2015
+:Post-History: XX
+:Requires: The new dependency-specification PEP.
+
+
+Abstract
+========
+
+This PEP specifies a programmatic interface for pip [#pip]_ and other
+distribution or installation tools to use when working with Python
+source trees (both the developer tree - e.g. the git tree - and source
+distributions).
+
+The programmatic interface allows decoupling of pip from its current
+hard dependency on setuptools [#setuptools]_ able for two
+key reasons:
+
+1. It enables new build systems that may be much easier to use without
+   requiring them to even appear to be setuptools.
+
+2. It facilitates setuptools itself changing its user interface without
+   breaking pip, giving looser coupling.
+
+The interface needed to permit pip to install build systems also enables pip to
+install build time requirements for packages which is an important step in
+getting pip to full feature parity with the installation components of
+easy-install.
+
+As PEP-426 [#pep426]_ is draft, we cannot utilise the metadata format it
+defined. However PEP-427 wheels are in wide use and fairly well specified, so
+we have adopted the METADATA format from that for specifying distribution
+dependencies and general project metadata. PEP-0508 [#pep508] provides a self
+contained language for describing a dependency, which we encapsulate in a thin
+JSON schema to describe bootstrap dependencies.
+
+Motivation
+==========
+
+There is significant pent-up frustration in the Python packaging ecosystem
+around the current lock-in between build system and pip. Breaking that lock-in
+is better for pip, for setuptools, and for other build systems like flit
+[#flit]_.
+
+Specification
+=============
+
+Overview
+--------
+
+Build tools will be located by reading a file ``pypa.json`` from the root
+directory of the source tree. That file describes how to get the build tool
+and the name of the command to run to invoke the tool.
+
+All tools will be expected to conform to a single command line interface
+modelled on pip's existing use of the setuptools setup.py interface.
+
+pypa.json
+---------
+
+The file ``pypa.json`` acts as a neutral configuration file for pip and other
+tools that want to build source trees to consult for configuration. The
+absence of a ``pypa.json`` file in a Python source tree implies a setuptools
+or setuptools compatible build system.
+
+The JSON has the following schema. Extra keys are ignored, which permits the
+use of ``pypa.json`` as a configuration file for other related tools. If doing
+that the chosen keys must be namespaced - e.g. ``flit`` with keys under that
+rather than (say) ``build`` or other generic keys.
+
+schema
+    The version of the schema. This PEP defines version "1".  Defaults to "1"
+    when absent. All tools reading the file must error on an unrecognised
+    schema version.
+
+bootstrap_requires
+    Optional list of dependency specifications [#dependencyspec] that must be
+    installed before running the build tool. For instance, if using flit, then
+    the requirements might be::
+
+        bootstrap_requires: ["flit"]
+
+build_command
+    A mandatory key, this is a list of Python format strings [#strformat]_
+    describing the command to run. For instance, if using flit then the build
+    command might be::
+
+        build_command: ["flit"]
+
+    If using a command which is a runnable module fred::
+
+        build_command: ["{PYTHON}", "-m", "fred"]
+
+Process interface
+-----------------
+
+The command to run is defined by a simple Python format string [#strformat]_.
+
+This permits build systems with dedicated scripts and those that are invoked
+using "python -m somemodule".
+
+Processes will be run with the current working directory set to the root of
+the source tree.
+
+When run, processes should not read from stdin - while pip currently runs
+build systems with stdin connected to it's own stdin, stdout and stderr are
+redirected and no communication with the user is possible.
+
+As usual with processes, a non-zero exit status indicates an error.
+
+Available format variables
+--------------------------
+
+PYTHON
+    The Python interpreter in use. This is important to enable calling things
+    which are just Python entry points.
+
+        {PYTHON} -m foo
+
+Available environment variables
+-------------------------------
+
+These variables are set by the caller of the build system and will always be
+available.
+
+PATH
+    The standard system path.
+
+PYTHON
+    As for format variables.
+
+PYTHONPATH
+    Used to control sys.path per the normal Python mechanisms.
+
+Subcommands
+-----------
+
+There are a number of separate subcommands that build systems must support.
+The examples below use a build_command of ``flit`` for illustrative purposes.
+
+build_requires
+    Query build requirements. Build requirements are returned as a UTF-8
+    encoded JSON document with one key ``build_requires`` consisting of a list
+    of dependency specifications [#dependencyspec]_. Additional keys must be
+    ignored. The build_requires command is the only command run without
+    setting up a build environment.
+
+    Example command::
+
+        flit build_requires
+
+metadata
+    Query project metadata.  The metadata and only the metadata should
+    be output on stdout in UTF-8 encoding. pip would run metadata just once to
+    determine what other packages need to be downloaded and installed. The
+    metadata is output as a wheel METADATA file per PEP-427 [#pep427]_.
+
+    Note that the metadata generated by the metadata command, and the metadata
+    present in a generated wheel must be identical.
+
+    Example command::
+
+        flit metadata
+
+wheel -d OUTPUT_DIR
+    Command to run to build a wheel of the project. OUTPUT_DIR will point to
+    an existing directory where the wheel should be output. Stdout and stderr
+    have no semantic meaning. Only one file should be output - if more are
+    output then pip would pick an arbitrary one to consume.
+
+    Example command::
+
+        flit wheel -d /tmp/pip-build_1234
+
+develop [--prefix PREFIX]
+    Command to do an in-place 'development' installation of the project.
+    Stdout and stderr have no semantic meaning.
+
+    Not all build systems will be able to perform develop installs. If a build
+    system cannot do develop installs, then it should error when run. Note
+    that doing so will cause use operations like ``pip install -e foo`` to
+    fail.
+
+    The prefix option is used for defining an alternative prefix for the
+    installation. While setuptools has ``--root`` and ``--user`` options,
+    they can be done equivalently using ``--prefix``, and pip or other
+    tools that accept ``--root`` or ``--user`` options should translate
+    appropriately.
+
+    The root option is used to define an alternative root within which the
+    command should operate.
+
+    For instance::
+
+        flit develop --root /tmp/ --prefix /usr/local
+
+    Should install scripts within `/tmp/usr/local/bin`, even if the Python
+    environment in use reports that the sys.prefix is `/usr/` which would lead
+    to using `/tmp/usr/bin/`. Similar logic applies for package files etc.
+
+The build environment
+---------------------
+
+Except for the build_requires command, all commands are run within a build
+environment. No specific implementation is required, but a build environment
+must achieve the following requirements.
+
+1. All dependencies specified by the project's build_requires must be
+   available for import from within ``$PYTHON``.
+
+1. All command-line scripts provided by the build-required packages must be
+   present in ``$PATH``.
+
+A corollary of this is that build systems cannot assume access to any Python
+package that is not declared as a build_requires or in the Python standard
+library.
+
+Hermetic builds
+---------------
+
+This specification does not prescribe whether builds should be hermetic or not.
+Existing build tools like setuptools will use installed versions of build time
+requirements (e.g. setuptools_scm) and only install other versions on version
+conflicts or missing dependencies. However its likely that better consistency
+can be created by always isolation builds and using only the specified dependencies.
+
+However there are nuanced problems there - such as how can users force the
+avoidance of a bad version of a build requirement which meets some packages
+dependencies. Future PEPs may tackle this problem, but it is not currently in
+scope - it does not affect the metadata required to coordinate between build
+systems and things that need to do builds, and thus is not PEP material.
+
+Upgrades
+--------
+
+'pypa.json' is versioned to permit future changes without requiring
+compatibility.
+
+The sequence for upgrading either of schemas in a new PEP will be:
+
+1. Issue new PEP defining an updated schema. If the schema is not entirely
+   backward compatible then a new version number must be defined.
+2. Consumers (e.g. pip) implement support for the new schema version.
+3. Package authors opt into the new schema when they are happy to introduce a
+   dependency on the version of 'pip' (and potentially other consumers) that
+   introduced support for the new schema version.
+
+The *same* process will take place for the initial deployment of this PEP:-
+the propogation of the capability to use this PEP without a `setuptools shim`_
+will be largely gated by the adoption rate of the first version of pip that
+supports it.
+
+Static metadata in sdists
+-------------------------
+
+This PEP does not tackle the current inability to trust static metadata in
+sdists. That is a separate problem to identifying and consuming the build
+system that is in use in a source tree, whether it came from an sdist or not.
+
+Handling of compiler options
+----------------------------
+
+Handling of different compiler options is out of scope for this specification.
+
+pip currently handles compiler options by appending user supplied strings to
+the command line it runs when running setuptools. This approach is sufficient
+to work with the build system interface defined in this PEP, with the
+exception that globally specified options will stop working globally as
+different build systems evolve. That problem can be solved in pip (or conda or
+other installers) without affecting interoperability.
+
+In the long term, wheels should be able to express the difference between
+wheels built with one compiler or options vs another, and that is PEP
+material.
+
+Examples
+========
+
+An example 'pypa.json' for using flit::
+
+  {"bootstrap_requires": ["flit"],
+   "build_command": "flit"}
+
+When 'pip' reads this it would prepare an environment with flit in it before
+trying to use flit.
+
+Because flit doesn't have setup-requires support today,
+`flit build_requires` would just output a constant string::
+
+  {"build_requires": []}
+
+`flit metadata` would interrogate `flit.ini` and marshal the metadata into
+a wheel METADATA file and output that on stdout.
+
+`flit wheel` would need to accept a `-d` parameter that tells it where to output the
+wheel (pip needs this).
+
+Backwards Compatibility
+=======================
+
+Older pips will remain unable to handle alternative build systems.
+This is no worse than the status quo - and individual build system
+projects can decide whether to include a shim ``setup.py`` or not.
+
+All existing build systems that can product wheels and do develop installs
+should be able to run under this abstraction and will only need a specific
+adapter for them constructed and published on PyPI.
+
+In the absence of a ``pypa.json`` file, tools like pip should assume a
+setuptools build system and use setuptools commands directly.
+
+Network effects
+---------------
+
+Projects that adopt build systems that are not setuptools compatible - that
+is that they have no setup.py, or the setup.py doesn't accept commands that
+existing tools try to use - will not be installable by those existing tools.
+
+Where those projects are used by other projects, this effect will cascade.
+
+In particular, because pip does not handle setup-requires today, any project
+(A) that adopts a setuptools-incompatible build system and is consumed as a
+setup-requirement by a second project (B) which has not itself transitioned to
+having a pypa.json will make B uninstallable by any version of pip. This is
+because setup.py in B will trigger easy-install when 'setup.py egg_info' is
+run by pip, and that will try and fail to install A.
+
+As such we recommend that tools which are currently used as setup-requires
+either ensure that they keep a `setuptools shim`_ or find their consumers and
+get them all to upgrade to the use of a `pypa.json` in advance of moving
+themselves. Pragmatically that is impossible, so the advice is to keep a
+setuptools shim indefinitely - both for projects like pbr, setuptools_scm and
+also projects like numpy.
+
+setuptools shim
+---------------
+
+It would be possible to write a generic setuptools shim that looks like
+``setup.py`` and under the hood uses ``pypa.json`` to drive the builds. This
+is not needed for pip to use the system, but would allow package authors to
+use the new features while still retaining compatibility with older pip
+versions.
+
+Rationale
+=========
+
+This PEP started with a long mailing list thread on distutils-sig [#thread]_.
+Subsequent to that a online meeting was held to debug all the positions folk
+had.  Minutes from that were posted to the list [#minutes]_.
+
+This specification is a translation of the consensus reached there into PEP
+form, along with some arbitrary choices on the minor remaining questions.
+
+The basic heuristic for the design has been to focus on introducing an
+abstraction without requiring development not strictly tied to the
+abstraction. Where the gap is small to improvements, or the cost of using the
+existing interface is very high, then we've taken on having the improvement as
+a dependency, but otherwise defered such to future iterations.
+
+We chose wheel METADATA files rather than defining a new specification,
+because pip can already handle wheel .dist-info directories which encode all
+the necessary data in a METADATA file. PEP-426 can't be used as it's still
+draft, and defining a new metadata format, while we should do that, is a
+separate problem. Using a directory on disk would not add any value to the
+interface (pip has to do that today due to limitations in the setuptools
+CLI).
+
+The use of 'develop' as a command is because there is no PEP specifying the
+interoperability of things that do what 'setuptools develop' does - so we'll
+need to define that before pip can take on the responsibility for doing the
+'develop' step. Once that's done we can issue a successor PEP to this one.
+
+The use of a command line API rather than a Python API is a little
+contentious. Fundamentally anything can be made to work, and the pip
+maintainers have spoken strongly in favour of retaining a process based
+interface - something that is mature and robust in pip today.
+
+The choice of JSON as a file format is a compromise between several
+constraints. Firstly there is no stdlib YAML interpreter, nor one for any of
+the other low-friction structured file formats. Secondly, INIParser is a poor
+format for a number of reasons, primarily that it has very minimal structure -
+but pip's maintainers are not fond of it. JSON is in the stdlib, has
+sufficient structure to permit embedding anything we want in future without
+requiring embedded DSL's.
+
+Donald suggested using ``setup.cfg`` and the existing setuptools command line
+rather than inventing something new. While that would permit interoperability
+with less visible changes, it requires nearly as much engineering on the pip
+side - looking for the new key in setup.cfg, implementing the non-installed
+environments to run the build in. And the desire from other build system
+authors not to confuse their users by delivering something that looks like but
+behaves quite differently to setuptools seems like a bigger issue than pip
+learning how to invoke a custom build tool.
+
+The metadata and wheel commands are required to have consistent metadata to
+avoid a race condition that could otherwise happen where pip reads the
+metadata, acts on it, and then the resulting wheel has incompatible
+requirements. That race is exploited today by packages using PEP-426
+environment markers, to work with older pip versions that do not support
+environment markers. That exploit is not needed with this PEP, because either
+the setuptools shim is in use (with older pip versions), or an environment
+marker ready pip is in use. The setuptools shim can take care of exploiting
+the difference older pip versions require.
+
+We discussed having an sdist verb. The main driver for this was to make sure
+that build systems were able to produce sdists that pip can build - but this is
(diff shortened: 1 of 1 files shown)


rgommers:
You may want to edit the title of this PR to reflect its content.

build-system-abstraction.rst
```diff
@@ -0,0 +1,360 @@
...
+This PEP specifies a programmatic interface for pip [#pip]_ and other
+distribution or installation tools to use when working with Python
+source trees (both the developer tree - e.g. the git tree - and source
+distributions).
+
+The programmatic interface allows decoupling of pip from its current
+hard dependency on setuptools [#setuptools]_ able for two
+key reasons:
+
+1. It enables new boutique build systems that may be much easier to use
```
rgommers: "new boutique" doesn't sound right. "boutique" says to me "small/niche/specialized", while I'd hope to see something that becomes widely used/applicable fast. And there are also existing build systems. Maybe rephrase?

build-system-abstraction.rst
```diff
@@ -0,0 +1,360 @@
...
+   without requiring them to even appear to be setuptools.
+
+2. It facilitates setuptools itself changing it's user interface without
```
rgommers: it's -> its

build-system-abstraction.rst
```diff
@@ -0,0 +1,360 @@
...
+   breaking pip, giving looser coupling.
+
+The programmatic interface also enables pip to install build time requirements
+for packages which is an important step in getting pip to full feature parity
+with the instalation components of easy-install.
```
rgommers: ``easy_install``

build-system-abstraction.rst
```diff
@@ -0,0 +1,360 @@
...
+pypa.yaml
+---------
+
+The yaml file has the following schema. Extra keys are ignored.
+
+schema
+    The version of the schema. This PEP defines version 1.
+    Defaults to 1
+bootstrap-requires
+    The requirements required to run the build tool itself.
```
rgommers: This sentence really doesn't parse and was the main source of my confusion. ``The requirements required to run the build tool`` is not the same as ``The build tool (interface)``. In your example this was simply ``flit``, while I read it as "the dependencies to be able to run flit.

rgommers: And can you add to the description that this key can be empty, for example if the build tool isn't a Python package on PyPi?

build-system-abstraction.rst
```diff
@@ -0,0 +1,360 @@
...
+          - requirement[extra]
+          - requirement:python_version<"3"
+          - requirement[extra]:python_version>"2"
+          - requirement[extra]>2.5:python_version=="2.7"
+
+build-tool
+    A command to run to query the build tool for its complete interface.
+    The build tool should output on stdout a build tool description JSON document.
+    Stdin may not be read from, and stderr may be handled however the calling process
+    desires.
```
rgommers: Examples:

    flit --dump-build-description
    make self-describe

Not sure if ``make``, ``cMake`` & co meet the stdin/stdout/stderr requirements here. That would be desirable, right?

build-system-abstraction.rst
```diff
@@ -0,0 +1,360 @@
...
+    of the calling process. The build-requires for the project will be
+    present in the Python environment when the develop command is run.
+    
+    Not all build systems will be able to perform develop installs. If a build
+    system cannot do develop installs, then this key can be omitted.
+    
+    Note tht when it is omitted, commands like ``pip install -e foo``
+    will be unable to complete.
+provided-by
+    The distribution name that define this build system interface.
```
rgommers: typo: define --> defines

build-system-abstraction.rst
```diff
@@ -0,0 +1,360 @@
...
+    This is used to facilitate caching the build tool output to avoid requiring
```
rgommers: I had to read this three times. ``build tool output`` here means the output of running the entry under ``build-tool`` from ``pypa.yaml``. (it could also have meant a wheel/sdist/... - a build tool outputs a lot of things)

build-system-abstraction.rst
rbtcollins: Well, if you can suggest something. But by definition anything less comprehensive than setuptools is boutique! I'll drop the word boutique.

build-system-abstraction.rst
rbtcollins: done

build-system-abstraction.rst
rbtcollins: Why? I'm naming the thing, not specifying the CLI script or the module.

build-system-abstraction.rst
rbtcollins: If its not a Python package on PyPI, it won't be installable and thus won't work. Can you enlarge on what you're thinking here?

rbtcollins: (Perhaps you're thinking of globally installed build systems or some such - I guess thats reasonable, but as its defined as a list empty is implicitly ok so I'm not sure we need to do anything).

build-system-abstraction.rst
rbtcollins: They can (they pass stdin to child processes, so its possibly to meet, or not meet the requirements)

build-system-abstraction.rst
rbtcollins: done

build-system-abstraction.rst
rbtcollins: rephrased

build-system-abstraction.rst
rgommers: I thought the thing is also named with an underscore. At least, that's how it's used on https://pypi.python.org/pypi/setuptools

build-system-abstraction.rst
rgommers: OK. Yes, I was indeed thinking about something already installed. Even for example ``waf`` as a Python-based build system is not pip-installable.

build-system-abstraction.rst
rgommers: thanks, clear

build-system-abstraction.rst
rgommers: New text looks good to me.

build-system-abstraction.rst
```diff
@@ -0,0 +1,419 @@
...
+build-requires
+    Command to run to query build requirements. Build requirements are
+    returned as a JSON document with one key ``build_requires`` consisting of
+    a list of dependency specifications. Additional keys must be ignored.
+
+metadata
+    Command to run to generate metadata for the project.  The metadata should
+    be output on stdout, stdin may not be consumed, and stderr handling is up
+    to the discretion of the calling process.  The build-requires for the
+    project will be present in the Python environment when the metadata
```
rgommers: Is this necessary? If I read this right it's very expensive. For example for projects that use the numpy C API, numpy itself has to be built/installed before you can get at the metadata for the project.

build-system-abstraction.rst
```diff
@@ -0,0 +1,419 @@
...
+-------------------------
+
+This PEP does not tackle the current inability to trust static metadata in
+sdists. That is a separate problem to identifying and consuming the build
+system that is in use in a source tree, whether it came from an sdist or not.
+
+Handling of compiler options
+----------------------------
+
+Handling of different compiler options is out of scope for this specification.
```
rgommers: What you'd want to pass to setuptools here is stuff like ``config --compiler=mingw32``. Is that what you mean by "compiler options"? I initially read it as "compiler flags".

rgommers: Maybe "build options" in that case?

rgommers: And be explicit below that they're only passed to the ``wheel`` and ``develop`` commands, not ``metadata``.

build-system-abstraction.rst
rgommers: The new version of the text is clear to me, thanks.

build-system-abstraction.rst
rgommers: text is quite clear now.

build-system-abstraction.rst
```diff
@@ -0,0 +1,419 @@
...
+the necessary data in a METADATA file. PEP-426 can't be used as it's still
+draft, and defining a new metadata format, while we should do that, is a
+separate problem. Using a directory on disk would not add any value to the
+interface (pip has to do that today due to limitations in the setuptools
+CLI).
+
+The use of 'develop' as a command is because there is no PEP specifying the
+interoperability of things that do what 'setuptools develop' does - so we'll
+need to define that before pip can take on the responsibility for doing the
+'develop' step. Once thats done we can issue a successor PEP to this one.
```
rgommers: typo ``thats`` --> ``that is``

rgommers:
This version looks quite good and clear to me.

build-system-abstraction.rst
```diff
@@ -0,0 +1,419 @@
...
+
+OUTPUT_DIR
+    Where to create requested output. If not set, outputting to the current
+    working directory is appropriate. This is required by ``pip`` which needs
+    to have wheels output in a known location.
+
+pypa.json
+---------
+
+The yaml file has the following schema. Extra keys are ignored.
```
xavfernandez: yaml -> json

build-system-abstraction.rst
```diff
@@ -0,0 +1,419 @@
...
+'pypa.json' has a schema version that they do not recognise.
+
+Build tools listed in a 'pypa.json' with schema version 1 must not generate a
+build tool description with a version other than 1 (or absent, as the default
+is 1).
+
+Thus the sequence for upgrading either of schemas in a new PEP will be:
+
+1. Issue new PEP defining build tool description and 'pypa.json' schemas. The
+   'pypa.yaml' schema version must change if either the 'pypa.yaml' schema has
```
xavfernandez: yaml-> json

build-system-abstraction.rst
```diff
@@ -0,0 +1,419 @@
...
+   changed or the build tool description schema has changed. The build tool
+   description schema version must change if the build tool description schema
+   has changed.
+2. Consumers (e.g. pip) implement support for the new schema version.
+3. Build tool authors implement the new schemas, and publish updated reference
+   'pypa.json' files for their users.  4. Package authors opt into the new
```
xavfernandez: Missing a newline for 4.

build-system-abstraction.rst
```diff
@@ -0,0 +1,419 @@
...
+=========
+
+This PEP started with a long mailing list thread on distutils-sig [#thread]_.
+Subsequent to that a online meeting was held to debug all the positions folk
+had.  Minutes from that were posted to the list [#minutes]_.
+
+This specification is a translation of the consensus reached there into PEP
+form, along with some arbitrary choices on the minor remaining questions.
+
+The basic heuristic for the design has to been to focus on introducing an
```
xavfernandez: has been to

build-system-abstraction.rst
rbtcollins: I think we'd pass them to metadata too, since they may affect the metadata.

rbtcollins: Further, the whole point of them being out of scope is that we're not defining their behaviour: we're saying that tools like conda or pip that choose to invoke build systems can build their own UI and logic around this, we're merely calling out that this is likely going to happen, and that what is passed where needs to understand different build systems will have different things here.

build-system-abstraction.rst
rbtcollins: done

build-system-abstraction.rst
```diff
@@ -0,0 +1,456 @@
...
+    present in a generated wheel must be identical.
+
+    Example command::
+
+        flit metadata
+
+wheel -d OUTPUT_DIR
+    Command to run to build a wheel of the project. OUTPUT_DIR will point to
+    an existing directory where the wheel should be output. Stdout and stderr
+    have no semantic meaning. Only one file should be output - if more are
```
dstufft: I'm not sure if it makes sense to indicate it here, but for instance pip will display stdout/stderr if the build fails, providing an interface for the build command to indicate why a build failed to the end user?

build-system-abstraction.rst
```diff
@@ -0,0 +1,456 @@
...
+
+Except for the build_requires command, all commands are run within a build
+environment. No specific implementation is required, but a build environment
+must achieve the following requirements.
+
+1. All dependencies specified by the project's build_requires must be
+   available for import from within ``$PYTHON``.
+
+1. All command-line scripts provided by the build-required packages must be
+   present in ``$PATH``.
```
dstufft: Should ``$PATH`` be mentioned in the environment variables that are provided up above?

build-system-abstraction.rst
```diff
@@ -0,0 +1,456 @@
...
+interoperability of things that do what 'setuptools develop' does - so we'll
+need to define that before pip can take on the responsibility for doing the
+'develop' step. Once that's done we can issue a successor PEP to this one.
+
+The use of a command line API rather than a Python API is a little
+contentious. Fundamentally anything can be made to work, and the pip
+maintainers have spoken strongly in favour of retaining a process based
+interface - something that is mature and robust in pip today.
+
+The choice of JSON as a file format is a compromise between several
```
dstufft: As much as I don't like TOML as a format, I think we should at least consider a ``pypa.toml`` file here (ala ``cargo.toml`` in Rust). It's nicer to write by hand than JSON (which has no comments, and is strict about things like trailing comma) and I see several single source, pure python libraries available for it in a quick search.

dstufft:
https://mail.python.org/pipermail/distutils-sig/2016-January/028152.html

build-system-abstraction.rst
rbtcollins: If there was semantic meaning, I'd be expecting something machine consumable. I think its fine for build invokers to choose to show stdout/err to users - or to hide it; I don't think we should be defining UI of such tools here.

build-system-abstraction.rst
rbtcollins: Hmm, sure.

build-system-abstraction.rst
dstufft: Oh, maybe I just had a different thinking of what semantic meant :] I was just thinking suggesting that you should put build errors out on one of those two.

build-system-abstraction.rst
rbtcollins: I'm recusing myself from choice here: I fundamentally don't care. Its the least interesting thing about the format. Pick something you like, and off we go.

If I was arguing, I'd argue for JSON at this point, since you don't like TOML as a format, and the stuff we're putting in here is very easy to hand-write in JSON - no harder than writing Python code TBH.

xavfernandez: Doesn't TOML suffer from the same weakness as YAML , i.e. the lack of stdlib interpreter ?

build-system-abstraction.rst
```diff
@@ -0,0 +1,456 @@
...
+directory of the source tree. That file describes how to get the build tool
+and the name of the command to run to invoke the tool.
+
+All tools will be expected to conform to a single command line interface
+modelled on pip's existing use of the setuptools setup.py interface.
+
+pypa.json
+---------
+
+The file ``pypa.json`` acts as neutron configuration file for pip and other
```
sigmavirus24: neutron?

rbtcollins: bwhah. will change to neutral.

build-system-abstraction.rst
takluyver: It does, but it's a much simpler format, and the primary module for it is pure python, whereas PyYAML tries to build an extension module.

Personally, I do like toml as a format - I'm considering transitioning to toml config files for flit. But these metadata files, as currently proposed, are so small and rarely changing that I don't think it's a problem to work with them in JSON.

If we do consider `pypa.toml`, though, I'd like to suggest a namespace which build tools can use to store their own metadata, to reduce the clutter of separate files a bit:

```toml
schema = 1
bootstrap_requires = ["flit"]
build_command = ["flit"]

[build.flit]
module = "foobar"
# ...
```

The build system would be responsible for reading the file again if it wanted to use this, and there would be no requirement for build tools to use this file at all.

This would be more complexity than I'd want to put in a JSON file, though.

build-system-abstraction.rst
```diff
@@ -0,0 +1,456 @@
...
+    Command to run to build a wheel of the project. OUTPUT_DIR will point to
+    an existing directory where the wheel should be output. Stdout and stderr
+    have no semantic meaning. Only one file should be output - if more are
+    output then pip would pick an arbitrary one to consume.
+
+    Example command::
+
+        flit wheel -d /tmp/pip-build_1234
+
+develop [--prefix PREFIX] [--root ROOT]
```
takluyver: I recently added support in flit for installing to a Python other than the one running `flit` (takluyver/flit#72). This expects the path to a Python interpreter, which it runs in a subprocess to find the installation paths.

I don't think `--prefix` and `--root` are sufficient to deal with all of the possibilities. Looking at the install schemes in sysconfig:

- A `--user` install is in a different format, especially on Windows and on OSX framework builds, so it can't be achieved just by setting the prefix.
- On POSIX, there may be separate prefixes for purelib and platlib (doesn't affect flit, since it only deals with purelib)

At a minimum, I think this needs an added `--user` flag, and ? some way of indicating the use of an OSX framework build (I don't really know what that is, just that there's a separate `osx_framework_user`scheme). Maybe it should pass the subprocess all relevant directories explicitly, e.g. in environment variables.

I also find the `--root` and `--prefix` distinction a bit confusing. Is there anything that should be installed within root but outside prefix? In the example below, prefix is an absolute path, so joining them the obvious way with `os.path.join` would just give `/usr/local`. Are absolute paths (including the drive) allowed on Windows, or just a leading `os.sep`? Can pip handle the concatenation and give me a single path?

takluyver:
Is there still interest in pushing this forwards?

rbtcollins:
certainly there is interest in pushing this forward. I don't know that we have to support all the possible interactions of ways of putting a Python layout together: at the heart of it, we're never going to be called for a different interpreter - pip always uses the interpreter of the resulting thing to invoke builds.

takluyver:
> I don't know that we have to support all the possible interactions of ways of putting a Python layout together: at the heart of it, we're never going to be called for a different interpreter - pip always uses the interpreter of the resulting thing to invoke builds.

Was this meant as a reply to [this comment of mine](https://github.com/pypa/interoperability-peps/pull/54#discussion_r52298754)? If so: pip itself will always be running in the target Python interpreter, but that need not be true of the build tool it's invoking for `pip install -e`. In the case of flit, for instance, it requires Python 3, so if you want to install into a Python 2.7 environment, flit will necessarily be running with another interpreter, and it needs to know where to install to.

We can also imagine build tools that need not run inside a Python interpreter at all.

Maybe development installs (`-e`) are too complex a topic, and we should leave them out of this proposal and write a separate proposal to standardise that interface.

build-system-abstraction.rst
vsajip: +1 for JSON at this point. Personal preferences should take second place to in-stdlib support, IMO.

If possible, a standardised dictionary schema should be identified at an API level, which might obviate any bikeshedding on file formats.

njsmith:
Wait, this py3 thing is a problem :-(

Even without editable mode, there's a problem with `pip install pkgname` (assuming no wheel is available, which I recognize is unlikely for flit-using packages), or `pip install path/` (much more likely, since this will be the standard incantation that we teach users for installing arbitrary packages from e.g. git checkouts, instead of `setup.py install`). These commands have to be run using the target python, simply because that's the only python we know about or that is guaranteed to exist -- random users don't necessarily have a python 3 interpreter available, and `pip install python3` isn't really a thing.

In a sense I guess this is not too different from how packages may also require that you provide C compiler and so forth. But in that case we still assume that the logic that provides glue between pip and the C compiler has to be pip installable, because what other choice do we have? I suppose you could make a python 2 compatible flit shim that just finds a copy of python3 and uses it to flit (or print a sensible error if this fails).

(Or I guess technically there's nothing stopping us from making wheels that contain a python 3 interpreter inside them, except good sense and/or a certain ingrained conformity to traditional norms.)

takluyver:
...I semi-seriously like the last suggestion. Provide 'Python 2' wheels for flit that bundle their own copy of a Python 3 interpreter. I'm probably not actually going to bother to do that, because messing around with all the different builds sounds tedious, but it's amusing to know that it's possible.

rgommers:
> But in that case we still assume that the logic that provides glue between pip and the C compiler has to be pip installable, because what other choice do we have?

That it's not pip-installable, which is perfectly reasonable. 

I think you're asking for major problems by trying to have non-wheel cross-talk between different Python installs. I'd suggest to stick to a simple rule: either the build tool uses the same interpreter, or it's managed by the user. 

> In the case of flit, for instance, it requires Python 3, so if you want to install into a Python 2.7 environment, flit will necessarily be running with another interpreter, and it needs to know where to install to.

``flit`` doesn't need to install anything, that's what ``pip`` is for. Even for a ``develop`` command, ``flit`` should build the project in-place, and ``pip`` should take care of installing the ``egg_info`` or similar file to make that in-place build visible right?

build-system-abstraction.rst
rbtcollins: Added a section on explicit namespace for cooperating tools.

build-system-abstraction.rst
rbtcollins: Discussion on IRC has concluded that we really only need --prefix, - --user is defined as --prefix=site.USER_BASE. --root is equivalent to --prefix ROOT+/path/to/existing/root.

e.g. --prefix foo
gets you foo/bin
--root foo
gets you foo/home/robertc/.virtualenvs/scratch/foo/bin

when using a virtualenv in /home/robertc/.virtualenvs/scratch

rbtcollins:
So python3 and python2 and flit. when pip is installing something in a python 2 environment there is no guarantee python3 is available - and pip won't know how to install flit in that situation - it would need to be treated like e.g. 'make' or 'autoconf'. If flit wants to be used for building python2 compatible things, and be like that, it certainly can. That doesn't affect the spec here though - just use the already defined variable PYTHON to pass the PYTHON that pip is actually using down for however flit's cross-python stuff is defined, and you should be good to go.

build-system-abstraction.rst
```diff
@@ -0,0 +1,472 @@
...
+environment markers. That exploit is not needed with this PEP, because either
+the setuptools shim is in use (with older pip versions), or an environment
+marker ready pip is in use. The setuptools shim can take care of exploiting
+the difference older pip versions require.
+
+We discussed having an sdist verb. The main driver for this was to make sure
+that build systems were able to produce sdists that pip can build - but this is
+circular: the whole point of this PEP is to let pip consume such sdists
+reliably and without requiring an implementation of setuptools. Further, while
+most everyone agrees that encouraging sdists to be uploaded to PyPI, there
```
rgommers: typo: ``encouraging ... "is good"`` or ``uploading sdists to PyPI is encouraged``.

takluyver:
@rgommers :

> Even for a develop command, flit should build the project in-place, and pip should take care of installing the egg_info or similar file to make that in-place build visible right?

Possibly, but that is not what this proposal currently says. It says that pip invokes the build tool to do a development install however the build tool wants.

I also don't like the 'egg link' method that setuptools/pip uses, so flit will retain the ability to do development installs using symlinks, even if pip doesn't invoke it.

build-system-abstraction.rst
```diff
@@ -0,0 +1,472 @@
...
+    Command to run to build a wheel of the project. OUTPUT_DIR will point to
+    an existing directory where the wheel should be output. Stdout and stderr
+    have no semantic meaning. Only one file should be output - if more are
+    output then pip would pick an arbitrary one to consume.
+
+    Example command::
+
+        flit wheel -d /tmp/pip-build_1234
+
+develop [--prefix PREFIX]
```
takluyver: The description below needs to be updated to remove references to `--root`.

I still don't think this is sufficient. `--user` is *not* always equivalent to setting prefix to `site.USER_BASE`. For instance, on Windows, normal prefix installs put packages in `{base}/Lib/site-packages`, but a user install puts them in e.g. `{base}/Python35/site-packages`.

It also doesn't address the `osx_framework_user` case, though maybe that's rare enough that it can be safely ignored?

rgommers:
> >    Even for a develop command, flit should build the project in-place, and pip should take care of installing the egg_info or similar file to make that in-place build visible right?

> Possibly, but that is not what this proposal currently says. It says that pip invokes the build tool to do a development install however the build tool wants.

Ah, I didn't see that. That is then completely out of line with the intention of the rest of this PEP. The build tool should stick to building, not installing. It will also give problems: if the build tool inserts some ``egg_info`` thingy or symlink or ... that ``pip`` doesn't know about, and then the user does a normal install, there are two installed versions and it's not clear which one is first on ``sys.path``. @rbtcollins can this be changed?

> I also don't like the 'egg link' method that setuptools/pip uses, so flit will retain the ability to do development installs using symlinks, even if pip doesn't invoke it.

Totally agree that ``egg_info``, ``.pth`` files, etc. all smell.

takluyver:
I'm inclined to say that we should drop development installs from this proposal for now, and come back to them separately. It feels like they're a pretty big topic on their own.

build-system-abstraction.rst
```diff
@@ -0,0 +1,472 @@
...
+---------
+
+The file ``pypa.json`` acts as neutral configuration file for pip and other
+tools that want to build source trees to consult for configuration. The
+absence of a ``pypa.json`` file in a Python source tree implies a setuptools
+or setuptools compatible build system.
+
+The JSON has the following schema. Extra keys are ignored, which permits the
+use of ``pypa.json`` as a configuration file for other related tools. If doing
+that the chosen keys must be namespaced - e.g. ``flit`` with keys under that
```
takluyver: If the format is JSON, I *don't* think we should do this. I was only suggesting it if we decided to switch to something like TOML/YAML. I'm very happy to leave it as JSON, but in this case, I wouldn't allow build tools to put extra data in this file. JSON is good precisely because the file is small and simple.

Also, I don't think we should allow build tools to add even one arbitrary key at the top level, because then there's no way to have an 'unknown key' error. If we do allow this, I think we should pick one top-level key e.g. `buildtool` and require that build tools put their data under that. But again, if it's JSON, let's not bother - I'd rather stick with a separate file.

build-system-abstraction.rst
rbtcollins: It seems reasonable to me that pip can take care of translating --user on windows or posix though. Can you link to the docs for the osx_framework_user case? I have no knowledge of it.

build-system-abstraction.rst
rbtcollins: I don't understand - toml/  yaml are precisely as extensible, schema wise, as json for our needs. Why is it a good idea for toml/yaml yet not json ?

If we push the extensions down a level, they can still stomp on each other. As far as stomping on PEP specified keys, it seems to me that it is indeed a risk, so sure, I'll move it down a level.

build-system-abstraction.rst
takluyver: But translate it to what? There aren't enough variables being passed on to the subprocess to represent the necessary information in any form.

I don't know if the osx_framework_user thing is documented anywhere, but it's an install scheme defined in the code:
https://hg.python.org/cpython/file/tip/Lib/sysconfig.py#l72

build-system-abstraction.rst
takluyver: JSON is not great for files we want humans to edit, because it doesn't allow comments. But everything is trade offs, and there is a JSON parser in the standard library, so I think JSON is OK here because this metadata file is very small and unlikely to change much. If I were to include all of the config for flit in this file, it would be well beyond the threshold where working with JSON directly gets annoying. So I suggested the extra namespace *only if* we decided to change the format.

Or to look at it another way: I'm the one who asked for the extra namespace, and I'm explicitly saying that I will not use it in a JSON file - I'll continue with a separate flit config file instead. So keep it simple, and leave it out.

(To be clear: I'm not advocating for a change in format, I think JSON is fine for this. But so long as it is JSON, let's keep it small and simple)

rbtcollins:
re: editable installs. They exist, pip supports them. We can either break the workflow of any *users* of a project that adopts this thing, or we can keep supporting them and in future do the hard work to standardise them. I prefer not to break users.

rgommers:
> re: editable installs. They exist, pip supports them. We can either break the workflow of any users of a project that adopts this thing, or we can keep supporting them and in future do the hard work to standardise them. I prefer not to break users.

Nothing needs to be broken for users. My point is that the text now says that the build tool should *install*, while it should imho be saying that the build tool should *build* and the install tool should *install* that build, i.e. make it visible via an egg-info file or symlink. 

I think that that's what you already meant, so I'm simply asking for a textual clarification. If that's not what you meant, then I don't understand the reason why you're letting the build tool take responsibility for installing something.

dstufft:
I'm going to merge this PR not to accept it, but so that it's managed inside of the repository and I'll give @rbtcollins commit bit. I'm also going to assign it a number.

dstufft:
Ok. This is now PEP 516.

takluyver:
> re: editable installs. They exist, pip supports them. We can either break the workflow of any users of a project that adopts this thing, or we can keep supporting them and in future do the hard work to standardise them. I prefer not to break users.

It's certainly nice if we can keep it working for users, but defining how it works seems to be a major difficulty with this PEP, and I'd rather it went through without a provision for development installs than got stuck trying to deal with that.

@rgommers:

> My point is that the text now says that the build tool should install, while it should imho be saying that the build tool should build and the install tool should install that build, i.e. make it visible via an egg-info file or symlink. 

If we do this, then we need to define what the build tool should build. In the normal install case, it builds a wheel, but that doesn't work for a development install. Probably the logical thing for it to build is a dist-info directory, but that's still somewhat ambiguous - e.g. dist-info inside a wheel is a bit different from the installed dist-info defined in PEP 376. Also, for instance, there's not much point in recording hashes of all the files in a development install, because the point is that they may change without reinstallation.

rgommers:
> If we do this, then we need to define what the build tool should build.

Only where, not what. The install tool doesn't have to, and must not, care what is built by the build tool.

> In the normal install case, it builds a wheel, but that doesn't work for a development install. Probably the logical thing for it to build is a dist-info directory, but that's still somewhat ambiguous

No, it builds only whatever is needed to make the package work completely in the location it's building the development build. So usually nothing but compiled extensions and generated Python files (``version.py`` for example). ``dist-info`` isn't needed for anything here.

> e.g. dist-info inside a wheel is a bit different from the installed dist-info defined in PEP 376. Also, for instance, there's not much point in recording hashes of all the files in a development install, because the point is that they may change without reinstallation.

Hashes are not needed for anything, and what pip needs is already covered by the PEP, this text:

    Note that the metadata generated by the metadata command, and the metadata
    present in a generated wheel must be identical.

njsmith:
@rgommers: editable installs need metadata so that `pip` and other tools can figure out that the install exists, can know what version it is (so that they know it can satisfy dependencies), and to support uninstall. All of these are a bit broken right now, but just removing the metadata entirely would be a regression. Also, editable installs may include scripts, which need to get installed onto PATH somehow. Also, in a perfect world they should do things like respect the platlib / purelib distinction, and support all the various file install locations that wheels do -- this doesn't actually happen right now, but if we sit down to do editable installs "right" then it probably needs to be sorted.

I started out with a similar opinion as you, but @rbtcollins convinced me that the swamp is much wider and deeper than that :-(

rgommers:
>  editable installs need metadata so that pip and other tools can figure out that the install exists, can know what version it is (so that they know it can satisfy dependencies), and to support uninstall.

Clearly yes, but that metadata is what is output by the ``metadata`` command in this PEP and needs to be put there by the installer right? I've just tried ``pip install -e . --user`` to verify what it does, which isn't that complex. A file ``packagename.egg-link`` goes into site-packages, and a ``package-name.egg-info/`` is put in ``.``. ``pip uninstall package`` then only removes ``packagename.egg-link``. 

> All of these are a bit broken right now, but just removing the metadata entirely would be a regression. 

No one suggests removing the metadata. It can be retrieved by the install tool with the ``metadata`` and put in the right place.

> editable installs may include scripts, which need to get installed onto PATH somehow.

Yes, that's probably the one thing that's necessarily a bit complicated.

> Also, in a perfect world they should do things like respect the platlib / purelib distinction

Why? Is there a usecase other than in-place builds that I'm missing? The [pip docs](https://pip.pypa.io/en/stable/reference/pip_install/?highlight=editable%20install#vcs-support) explain exactly nothing about what editable installs are for, so I'm just going with the only usecase I'm aware of.

> I started out with a similar opinion as you, but @rbtcollins convinced me that the swamp is much wider and deeper than that :-(

If there's a discussion somewhere that I missed, can you provide a link? I'd really like to understand the actual problem here, because it seems quite problematic to me to have a PEP that attempts to cleanly split build from install, and then has a ``develop`` command that let's the build tool do an install.

PR #56, opened 2015-11-06
Define dependency specifications formally.

rbtcollins:
[Review on Reviewable](https://reviewable.io/reviews/pypa/interoperability-peps/56)

### DIFF
--- a/dependency-specification.rst
+++ b/dependency-specification.rst
@@ -1,0 +1,561 @@
+:PEP: XX
+:Title: Dependency specification for Python Software Packages
+:Version: $Revision$
+:Last-Modified: $Date$
+:Author: Robert Collins <rbtcollins@hp.com>
+:BDFL-Delegate: Donald Stufft <donald@stufft.io>
+:Discussions-To: distutils-sig <distutils-sig@python.org>
+:Status: Draft
+:Type: Standards Track
+:Content-Type: text/x-rst
+:Created: 11-Nov-2015
+:Post-History: XX
+
+
+Abstract
+========
+
+This PEP specifies the language used to describe dependencies for packages.
+It draws a border at the edge of describing a single dependency - the
+different sorts of dependencies and when they should be installed is a higher
+level problem. The intent is to provide a building block for higher layer
+specifications.
+
+The job of a dependency is to enable tools like pip [#pip]_ to find the right
+package to install. Sometimes this is very loose - just specifying a name, and
+sometimes very specific - referring to a specific file to install. Sometimes
+dependencies are only relevant in one platform, or only some versions are
+acceptable, so the language permits describing all these cases.
+
+The language defined is a compact line based format which is already in
+widespread use in pip requirements files, though we do not specify the command
+line option handling that those files permit. There is one caveat - the
+URL reference form, specified in PEP-440 [#pep440]_ is not actually
+implemented in pip, but since PEP-440 is accepted, we use that format rather
+than pip's current native format.
+
+Motivation
+==========
+
+Any specification in the Python packaging ecosystem that needs to consume
+lists of dependencies needs to build on an approved PEP for such, but
+PEP-426 [#pep426]_ is mostly aspirational - and there are already existing
+implementations of the dependency specification which we can instead adopt.
+The existing implementations are battle proven and user friendly, so adopting
+them is arguably much better than approving an aspirational, unconsumed, format.
+
+Specification
+=============
+
+Examples
+--------
+
+All features of the language shown with a name based lookup::
+
+    requests [security,tests] >= 2.8.1, == 2.8.* ; python_version < "2.7.10"
+
+A minimal URL based lookup::
+
+    pip @ https://github.com/pypa/pip/archive/1.3.1.zip#sha1=da9234ee9982d4bbb3c72346a6de940a148ea686
+
+Concepts
+--------
+
+A dependency specification always specifies a distribution name. It may
+include extras, which expand the dependencies of the named distribution to
+enable optional features. The version installed can be controlled using
+version limits, or giving the URL to a specific artifact to install. Finally
+the dependency can be made conditional using environment markers.
+
+Grammar
+-------
+
+We first cover the grammar briefly and then drill into the semantics of each
+section later.
+
+A distribution specification is written in ASCII text. We use a parsley
+[#parsley]_ grammar to provide a precise grammar. It is expected that the
+specification will be embedded into a larger system which offers framing such
+as comments, multiple line support via continuations, or other such features.
+
+The full grammar including annotations to build a useful parse tree is
+included at the end of the PEP.
+
+Versions may be specified according to the PEP-440 [#pep440]_ rules. (Note:
+URI is defined in std-66 [#std66]_::
+
+    version_cmp   = wsp* '<' | '<=' | '!=' | '==' | '>=' | '>' | '~=' | '==='
+    version       = wsp* ( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+
+    version_one   = version_cmp version wsp*
+    version_many  = version_one (wsp* ',' version_one)*
+    versionspec   = ( '(' version_many ')' ) | version_many
+    urlspec       = '@' wsp* <URI_reference>
+
+Environment markers allow making a specification only take effect in some
+environments::
+
+    marker_op     = version_cmp | (wsp* 'in') | (wsp* 'not' wsp+ 'in')
+    python_str_c  = (wsp | letter | digit | '(' | ')' | '.' | '{' | '}' |
+                     '-' | '_' | '*' | '#' | ':' | ';' | ',' | '/' | '?' |
+                     '[' | ']' | '!' | '~' | '`' | '@' | '$' | '%' | '^' |
+                     '&' | '=' | '+' | '|' | '<' | '>' )
+    dquote        = '"'
+    squote        = '\\''
+    python_str    = (squote (python_str_c | dquote)* squote |
+                     dquote (python_str_c | squote)* dquote)
+    env_var       = ('python_version' | 'python_full_version' |
+                     'os_name' | 'sys_platform' | 'platform_release' |
+                     'platform_system' | 'platform_version' |
+                     'platform_machine' | 'python_implementation' |
+                     'implementation_name' | 'implementation_version' |
+                     'extra' # ONLY when defined by a containing layer
+                     )
+    marker_var    = wsp* (env_var | python_str)
+    marker_expr   = marker_var marker_op marker_var
+                  | wsp* '(' marker wsp* ')'
+    marker_and    = marker_expr wsp* 'and' marker_expr
+                  | marker_expr
+    marker_or     = marker_and wsp* 'or' marker_and
+                      | marker_and
+    marker        = marker_or
+    quoted_marker = ';' wsp* marker
+
+Optional components of a distribution may be specified using the extras
+field::
+
+    identifier    = letterOrDigit (
+                    letterOrDigit |
+                    (( letterOrDigit | '-' | '_' | '.')* letterOrDigit ) )*
+    name          = identifier
+    extras_list   = identifier (wsp* ',' wsp* identifier)*
+    extras        = '[' wsp* extras_list? wsp* ']'
+
+Giving us a rule for name based requirements::
+
+    name_req      = name wsp* extras? wsp* versionspec? wsp* quoted_marker?
+
+And a rule for direct reference specifications::
+
+    url_req       = name wsp* extras? wsp* urlspec wsp+ quoted_marker?
+
+Leading to the unified rule that can specify a dependency.::
+
+    specification = wsp* ( url_req | name_req ) wsp*
+
+Whitespace
+----------
+
+Non line-breaking whitespace is mostly optional with no semantic meaning. The
+sole exception is detecting the end of a URL requirement.
+
+Names
+-----
+
+Python distribution names are currently defined in PEP-345 [#pep345]_. Names
+act as the primary identifier for distributions. They are present in all
+dependency specifications, and are sufficient to be a specification on their
+own. However, PyPI places strict restrictions on names - they must match a
+case insensitive regex or they won't be accepted. Accordingly in this PEP we
+limit the acceptable values for identifiers to that regex. A full redefinition
+of name may take place in a future metadata PEP. The regex (run with
+re.IGNORECASE) is::
+
+    ^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$
+
+Extras
+------
+
+An extra is an optional part of a distribution. Distributions can specify as
+many extras as they wish, and each extra results in the declaration of
+additional dependencies of the distribution **when** the extra is used in a
+dependency specification. For instance::
+
+    requests[security]
+
+Extras union in the dependencies they define with the dependencies of the
+distribution they are attached to. The example above would result in requests
+being installed, and requests own dependencies, and also any dependencies that
+are listed in the "security" extra of requests.
+
+If multiple extras are listed, all the dependencies are unioned together.
+
+Versions
+--------
+
+See PEP-440 [#pep440]_ for more detail on both version numbers and version
+comparisons. Version specifications limit the versions of a distribution that
+can be used. They only apply to distributions looked up by name, rather than
+via a URL. Version comparison are also used in the markers feature. The
+optional brackets around a version are present for compatibility with PEP-345
+[#pep345]_ but should not be generated, only accepted.
+
+Environment Markers
+-------------------
+
+Environment markers allow a dependency specification to provide a rule that
+describes when the dependency should be used. For instance, consider a package
+that needs argparse. In Python 2.7 argparse is always present. On older Python
+versions it has to be installed as a dependency. This can be expressed as so::
+
+    argparse;python_version<"2.7"
+
+A marker expression evalutes to either True or False. When it evaluates to
+False, the dependency specification should be ignored.
+
+The marker language is a subset of Python itself, chosen for the ability to
+safely evaluate it without running arbitrary code that could become a security
+vulnerability. Markers were first standardised in PEP-345 [#pep345]_. This PEP
+fixes some issues that were observed in the design described in PEP-426
+[#pep426]_.
+
+Comparisons in marker expressions are typed by the comparison operator.  The
+<marker_op> operators that are not in <version_cmp> perform the same as they
+do for strings in Python. The <version_cmp> operators use the PEP-440
+[#pep440]_ version comparison rules when those are defined (that is when both
+sides have a valid version specifier). If there is no defined PEP-440
+behaviour and the operator exists in Python, then the operator falls back to
+the Python behaviour. Otherwise an error should be raised. e.g. the following
+will result in  errors::
+
+    "dog" ~= "fred"
+    python_version ~= "surprise"
+
+User supplied constants are always encoded as strings with either ``'`` or
+``"`` quote marks. Note that backslash escapes are not defined, but existing
+implementations do support them. They are not included in this
+specification because they add complexity and there is no observable need for
+them today. Similarly we do not define non-ASCII character support: all the
+runtime variables we are referencing are expected to be ASCII-only.
+
+The variables in the marker grammar such as "os_name" resolve to values looked
+up in the Python runtime. With the exception of "extra" all values are defined
+on all Python versions today - it is an error in the implementation of markers
+if a value is not defined.
+
+Unknown variables must raise an error rather than resulting in a comparison
+that evaluates to True or False.
+
+Variables whose value cannot be calculated on a given Python implementation
+should evaluate to ``0`` for versions, and an empty string for all other
+variables.
+
+The "extra" variable is special. It is used by wheels to signal which
+specifications apply to a given extra in the wheel ``METADATA`` file, but
+since the ``METADATA`` file is based on a draft version of PEP-426, there is
+no current specification for this. Regardless, outside of a context where this
+special handling is taking place, the "extra" variable should result in an
+error like all other unknown variables.
+
+.. list-table::
+   :header-rows: 1
+
+   * - Marker
+     - Python equivalent
+     - Sample values
+   * - ``os_name``
+     - ``os.name``
+     - ``posix``, ``java``
+   * - ``sys_platform``
+     - ``sys.platform``
+     - ``linux``, ``linux2``, ``darwin``, ``java1.8.0_51`` (note that "linux"
+       is from Python3 and "linux2" from Python2)
+   * - ``platform_machine``
+     - ``platform.machine()``
+     - ``x86_64``
+   * - ``python_implementation``
+     - ``platform.python_implementation()``
+     - ``CPython``, ``Jython``
+   * - ``platform_release``
+     - ``platform.release()``
+     - ``3.14.1-x86_64-linode39``, ``14.5.0``, ``1.8.0_51``
+   * - ``platform_system``
+     - ``platform.system()``
+     - ``Linux``, ``Windows``, ``Java``
+   * - ``platform_version``
+     - ``platform.version()``
+     - ``#1 SMP Fri Apr 25 13:07:35 EDT 2014``
+       ``Java HotSpot(TM) 64-Bit Server VM, 25.51-b03, Oracle Corporation``
+       ``Darwin Kernel Version 14.5.0: Wed Jul 29 02:18:53 PDT 2015; root:xnu-2782.40.9~2/RELEASE_X86_64``
+   * - ``python_version``
+     - ``platform.python_version()[:3]``
+     - ``3.4``, ``2.7``
+   * - ``python_full_version``
+     - ``platform.python_version()``
+     - ``3.4.0``, ``3.5.0b1``
+   * - ``implementation_name``
+     - ``sys.implementation.name``
+     - ``cpython``
+   * - ``implementation_version``
+     - see definition below
+     - ``3.4.0``, ``3.5.0b1``
+   * - ``extra``
+     - An error except when defined by the context interpreting the
+       specification.
+     - ``test``
+
+The ``implementation_version`` marker variable is derived from
+``sys.implementation.version``::
+
+    def format_full_version(info):
+        version = '{0.major}.{0.minor}.{0.micro}'.format(info)
+        kind = info.releaselevel
+        if kind != 'final':
+            version += kind[0] + str(info.serial)
+        return version
+
+    if hasattr(sys, 'implementation'):
+        implementation_version = format_full_version(sys.implementation.version)
+    else:
+        implementation_version = "0"
+
+Backwards Compatibility
+=======================
+
+Most of this PEP is already widely deployed and thus offers no compatibility
+concerns.
+
+There are however a few points where the PEP differs from the deployed base.
+
+Firstly, PEP-440 direct references haven't actually been deployed in the wild,
+but they were designed to be compatibly added, and there are no known
+obstacles to adding them to pip or other tools that consume the existing
+dependency metadata in distributions - particularly since they won't be
+permitted to be present in PyPI uploaded distributions anyway.
+
+Secondly, PEP-426 markers which have had some reasonable deployment,
+particularly in wheels and pip, will handle version comparisons with
+``python_version`` "2.7.10" differently. Specifically in 426 "2.7.10" is less
+than "2.7.9". This backward incompatibility is deliberate. We are also
+defining new operators - "~=" and "===", and new variables -
+``platform_release``, ``platform_system``, ``implementation_name``, and
+``implementation_version`` which are not present in older marker
+implementations. The variables will error on those implementations. Users of
+both features will need to make a judgement as to when support has become
+sufficiently widespread in the ecosystem that using them will not cause
+compatibility issues.
+
+Thirdly, PEP-345 required brackets around version specifiers. In order to
+accept PEP-345 dependency specifications, brackets are accepted, but they
+should not be generated.
+
+Rationale
+=========
+
+In order to move forward with any new PEPs that depend on environment markers,
+we needed a specification that included them in their modern form. This PEP
+brings together all the currently unspecified components into a specified
+form.
+
+The requirement specifier was adopted from the EBNF in the setuptools
+pkg_resources documentation, since we wish to avoid depending on a defacto, vs
+PEP specified, standard.
+
+Complete Grammar
+================
+
+The complete parsley grammar::
+
+    wsp           = ' ' | '\t'
+    version_cmp   = wsp* <'<' | '<=' | '!=' | '==' | '>=' | '>' | '~=' | '==='>
+    version       = wsp* <( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+>
+    version_one   = version_cmp:op version:v wsp* -> (op, v)
+    version_many  = version_one:v1 (wsp* ',' version_one)*:v2 -> [v1] + v2
+    versionspec   = ('(' version_many:v ')' ->v) | version_many
+    urlspec       = '@' wsp* <URI_reference>
+    marker_op     = version_cmp | (wsp* 'in') | (wsp* 'not' wsp+ 'in')
+    python_str_c  = (wsp | letter | digit | '(' | ')' | '.' | '{' | '}' |
+                     '-' | '_' | '*' | '#' | ':' | ';' | ',' | '/' | '?' |
+                     '[' | ']' | '!' | '~' | '`' | '@' | '$' | '%' | '^' |
+                     '&' | '=' | '+' | '|' | '<' | '>' )
+    dquote        = '"'
+    squote        = '\\''
+    python_str    = (squote <(python_str_c | dquote)*>:s squote |
+                     dquote <(python_str_c | squote)*>:s dquote) -> s
+    env_var       = ('python_version' | 'python_full_version' |
+                     'os_name' | 'sys_platform' | 'platform_release' |
+                     'platform_system' | 'platform_version' |
+                     'platform_machine' | 'python_implementation' |
+                     'implementation_name' | 'implementation_version' |
+                     'extra' # ONLY when defined by a containing layer
+                     ):varname -> lookup(varname)
+    marker_var    = wsp* (env_var | python_str)
+    marker_expr   = marker_var:l marker_op:o marker_var:r -> (o, l, r)
+                  | wsp* '(' marker:m wsp* ')' -> m
+    marker_and    = marker_expr:l wsp* 'and' marker_expr:r -> ('and', l, r)
+                  | marker_expr:m -> m
+    marker_or     = marker_and:l wsp* 'or' marker_and:r -> ('or', l, r)
+                      | marker_and:m -> m
+    marker        = marker_or
+    quoted_marker = ';' wsp* marker
+    identifier    = <letterOrDigit (
+                    letterOrDigit |
+                    (( letterOrDigit | '-' | '_' | '.')* letterOrDigit ) )*>
+    name          = identifier
+    extras_list   = identifier:i (wsp* ',' wsp* identifier)*:ids -> [i] + ids
+    extras        = '[' wsp* extras_list?:e wsp* ']' -> e
+    name_req      = (name:n wsp* extras?:e wsp* versionspec?:v wsp* quoted_marker?:m
+                     -> (n, e or [], v or [], m))
+    url_req       = (name:n wsp* extras?:e wsp* urlspec:v wsp+ quoted_marker?:m
+                     -> (n, e or [], v, m))
+    specification = wsp* ( url_req | name_req ):s wsp* -> s
+    # The result is a tuple - name, list-of-extras,
+    # list-of-version-constraints-or-a-url, marker-ast or None
+
+
+    URI_reference = <URI | relative_ref>
+    URI           = scheme ':' hier_part ('?' query )? ( '#' fragment)?
+    hier_part     = ('//' authority path_abempty) | path_absolute | path_rootless | path_empty
+    absolute_URI  = scheme ':' hier_part ( '?' query )?
+    relative_ref  = relative_part ( '?' query )? ( '#' fragment )?
+    relative_part = '//' authority path_abempty | path_absolute | path_noscheme | path_empty
+    scheme        = letter ( letter | digit | '+' | '-' | '.')*
+    authority     = ( userinfo '@' )? host ( ':' port )?
+    userinfo      = ( unreserved | pct_encoded | sub_delims | ':')*
+    host          = IP_literal | IPv4address | reg_name
+    port          = digit*
+    IP_literal    = '[' ( IPv6address | IPvFuture) ']'
+    IPvFuture     = 'v' hexdig+ '.' ( unreserved | sub_delims | ':')+
+    IPv6address   = (
+                      ( h16 ':'){6} ls32
+                      | '::' ( h16 ':'){5} ls32
+                      | ( h16 )?  '::' ( h16 ':'){4} ls32
+                      | ( ( h16 ':')? h16 )? '::' ( h16 ':'){3} ls32
+                      | ( ( h16 ':'){0,2} h16 )? '::' ( h16 ':'){2} ls32
+                      | ( ( h16 ':'){0,3} h16 )? '::' h16 ':' ls32
+                      | ( ( h16 ':'){0,4} h16 )? '::' ls32
+                      | ( ( h16 ':'){0,5} h16 )? '::' h16
+                      | ( ( h16 ':'){0,6} h16 )? '::' )
+    h16           = hexdig{1,4}
+    ls32          = ( h16 ':' h16) | IPv4address
+    IPv4address   = dec_octet '.' dec_octet '.' dec_octet '.' Dec_octet
+    nz            = ~'0' digit
+    dec_octet     = (
+                      digit # 0-9
+                      | nz digit # 10-99
+                      | '1' digit{2} # 100-199
+                      | '2' ('0' | '1' | '2' | '3' | '4') digit # 200-249
+                      | '25' ('0' | '1' | '2' | '3' | '4' | '5') )# %250-255
+    reg_name = ( unreserved | pct_encoded | sub_delims)*
+    path = (
+            path_abempty # begins with '/' or is empty
+            | path_absolute # begins with '/' but not '//'
+            | path_noscheme # begins with a non-colon segment
+            | path_rootless # begins with a segment
+            | path_empty ) # zero characters
+    path_abempty  = ( '/' segment)*
+    path_absolute = '/' ( segment_nz ( '/' segment)* )?
+    path_noscheme = segment_nz_nc ( '/' segment)*
+    path_rootless = segment_nz ( '/' segment)*
+    path_empty    = pchar{0}
+    segment       = pchar*
+    segment_nz    = pchar+
+    segment_nz_nc = ( unreserved | pct_encoded | sub_delims | '@')+
+                    # non-zero-length segment without any colon ':'
+    pchar         = unreserved | pct_encoded | sub_delims | ':' | '@'
+    query         = ( pchar | '/' | '?')*
+    fragment      = ( pchar | '/' | '?')*
+    pct_encoded   = '%' hexdig
+    unreserved    = letter | digit | '-' | '.' | '_' | '~'
+    reserved      = gen_delims | sub_delims
+    gen_delims    = ':' | '/' | '?' | '#' | '(' | ')?' | '@'
+    sub_delims    = '!' | '$' | '&' | '\\'' | '(' | ')' | '*' | '+' | ',' | ';' | '='
+    hexdig        = digit | 'a' | 'A' | 'b' | 'B' | 'c' | 'C' | 'd' | 'D' | 'e' | 'E' | 'f' | 'F'
+
+A test program - if the grammar is in a string ``grammar``::
+
+    import os
+    import sys
+    import platform
+
+    from parsley import makeGrammar
+
+    grammar = """
+        wsp ...
+        """
+    tests = [
+        "A",
+        "aa",
+        "name",
+        "name>=3",
+        "name>=3,<2",
+        "name [fred,bar] @ http://foo.com ; python_version=='2.7'",
+        "name[quux, strange];python_version<'2.7' and platform_version=='2'",
+        "name; os_name=='a' or os_name=='b'",
+        # Should parse as (a and b) or c
+        "name; os_name=='a' and os_name=='b' or os_name=='c'",
+        # Overriding precedence -> a and (b or c)
+        "name; os_name=='a' and (os_name=='b' or os_name=='c')",
+        # should parse as a or (b and c)
+        "name; os_name=='a' or os_name=='b' and os_name=='c'",
+        # Overriding precedence -> (a or b) and c
+        "name; (os_name=='a' or os_name=='b') and os_name=='c'",
+        ]
+
+    def format_full_version(info):
+        version = '{0.major}.{0.minor}.{0.micro}'.format(info)
+        kind = info.releaselevel
+        if kind != 'final':
+            version += kind[0] + str(info.serial)
+        return version
+
+    if hasattr(sys, 'implementation'):
+        implementation_version = format_full_version(sys.implementation.version)
+        implementation_name = sys.implementation.name
+    else:
+        implementation_version = '0'
+        implementation_name = ''
+    bindings = {
+        'implementation_name': implementation_name,
+        'implementation_version': implementation_version,
+        'os_name': os.name,
+        'platform_machine': platform.machine(),
+        'platform_release': platform.release(),
+        'platform_system': platform.system(),
+        'platform_version': platform.version(),
+        'python_full_version': platform.python_version(),
+        'python_implementation': platform.python_implementation(),
+        'python_version': platform.python_version()[:3],
+        'sys_platform': sys.platform,
+    }
+
+    compiled = makeGrammar(grammar, {'lookup': bindings.__getitem__})
+    for test in tests:
+        parsed = compiled(test).specification()
+        print("%s -> %s" % (test, parsed))
+
+References
+==========
+
+.. [#pip] pip, the recommended installer for Python packages
+   (http://pip.readthedocs.org/en/stable/)
+
+.. [#pep345] PEP-345, Python distribution metadata version 1.2.
+   (https://www.python.org/dev/peps/pep-0345/)
+
+.. [#pep426] PEP-426, Python distribution metadata.
+   (https://www.python.org/dev/peps/pep-0426/)
+
+.. [#pep440] PEP-440, Python distribution metadata.
+   (https://www.python.org/dev/peps/pep-0440/)
+
+.. [#std66] The URL specification.
+   (https://tools.ietf.org/html/rfc3986)
+
+.. [#parsley] The parsley PEG library.
+   (https://pypi.python.org/pypi/parsley/)
+
+Copyright
+=========
+
+This document has been placed in the public domain.
+
+
+
+..
+   Local Variables:
+   mode: indented-text
+   indent-tabs-mode: nil
+   sentence-end-double-space: t
+   fill-column: 70
+   coding: utf-8
+   End:


dependency-specification.rst
```diff
@@ -0,0 +1,368 @@
...
+A dependency specification always specifies a distribution name. It may
+include extras, which expand the dependencies of the named distribution to
+enable optional features. The version installed can be controlled using
+version limits, or giving the URL to a specific artifact to install. Finally
+the dependency can be made conditional using environment markers.
+
+Grammar
+-------
+
+We first cover the grammar briefly and the drill into the semantics of each
```
xavfernandez: the drill -> then drill ?

dependency-specification.rst
```diff
@@ -0,0 +1,368 @@
+:PEP: XX
+:Title: Dependency specification for Python Software Packages
+:Version: $Revision$
+:Last-Modified: $Date$
+:Author: Robert Collins <rbtcollins@hp.com>
+BDFL-Delegate: Donald Stufft <donald@stufft.io>
```
merwok: Missing colon.

dependency-specification.rst
```diff
@@ -0,0 +1,368 @@
...
+.. list-table::
+   :header-rows: 1
+
+   * - Marker
+     - Python equivalent
+     - Sample values
+     - Default if unavailable
+   * - ``os_name``
+     - ``os.name``
+     - ``posix``, ``java``
```
merwok: Are users supposed to write `posix` or `"posix"` in their spec?

rbtcollins: This is showing the example values.

A marker expression using it would be something like
{{{
os_name=="java"
}}}

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
merwok: Is this made crystal clear in examples or somewhere else in the PEP?

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+Specification
+=============
+
+Examples
+--------
+
+All features of the language shown with a name based lookup::
+
+    requests \
+        [security] >= 2.8.1, == 2.8.* \
```
njsmith: Might be good to include multiple extras in the example to demonstrate the comma syntax

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+continuations - a single backslash followed by a new line ('\\\\n')::
+
+    CSP           = 1*(WSP / ("\\" LF))
+
+Versions may be specified according to the PEP-440 [#pep440]_ rules. (Note:
+URI is defined in std-66 [#std66]_::
+
+    version-cmp   = "<" / "<=" / "!=" / "==" / ">=" / ">" / "~=" / "==="
+    version       = 1*( DIGIT / ALPHA / "-" / "_" / "." / "*" )
+    versionspec   = ["("] version-cmp version *(',' version-cmp version) [")"]
```
njsmith: As written I believe this allows for unbalanced parentheses? Kinda fiddly but should probably be
```
version-spec-contents = version-cmp version *("," version-cmp version)
version-spec = version-spec-contents / "(" version-spec-contents ")"
```

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+    urlspec       = "@" URI
+
+Environment markers allow making a specification only take effect in some
+environments::
+
+    marker-op     = version-cmp / "in" / "not in"
+    python-str-c  = (WSP / ALPHA / DIGIT / "(" / ")" / "." / "{" / "}" /
+                    "-" / "_" / "*"
```
njsmith: Unbalanced parens

njsmith: Also, is it intentional that backslash escapes aren't supported? Possibly deserves a note if so.

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+    marker-var    = python-str / "python_version" / "python_full_version" /
+                    "os_name"" / "sys_platform" / "platform_release" /
+                    "platform_version" / "platform_machine" /
+                    "platform_python_implementation" / "implementation_name" /
+                    "implementation_version" / "platform_dist_name"
+                    "platform_dist_version" / "platform_dist_id"
+    marker-expr   = "(" marker ")" / (marker-var [marker-op marker-var])
+    marker        = marker-expr *( ("and" / "or") marker-expr)
+    name-marker   = ";" *CSP marker
+    url-marker    = ";" 1*CSP marker
```
njsmith: Why the difference between these two rules?

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+Optional components of a distribution may be specified using the extras
+field::
+
+    identifier    = 1*( DIGIT / ALPHA / "-" / "_" )
+    name          = identifier
+    extras        = "[" identifier *("," identifier) "]"
+
+Giving us a rule for name based requirements::
+
+    name_req      = name [CSP extras] [CSP versionspec] [CSP name-marker]
```
njsmith: AFAICT the existing requirements.txt does not require whitespace before `extras` (e.g. https://pip.readthedocs.org/en/latest/reference/pip_install/#requirement-specifiers has `SomeProject[foo, bar]` as an example).

It's also unclear to me whether the existing rule is actually that extras have to come before versionspec (though I think this actually would be a good rule!). The [old docs](https://pip.readthedocs.org/en/1.1/requirements.html#the-requirements-file-format) has an example requirements.txt line of: `MyPackage==3.0 [PDF]`.

njsmith: Oh, and it definitely doesn't require any whitespace between the name and the versionspec...

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+
+And a rule for direct reference specifications::
+
+    url_req       = name [CSP extras] urlspec [CSP url-marker]
```
njsmith: Similar concerns here.

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+
+Leading to the unified rule that can specify a dependency::
+
+    specification = (name_req / location_req) [CSP comment]
```
njsmith: Requiring whitespace before a comment marker seems... well, not *wrong* exactly, but still surprisingly strict.

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+
+Whitespace
+----------
+
+Non line-breaking whitespace is optional and has no semantic meaning.
```
njsmith: Ah, well, that changes things :-).

If all whitespace is optional, then why is `CSP` specifically defined as "at least one whitespace character" instead of "zero or more whitespace characters", and why do `name-marker` and `url-marker` differ only in that the latter has required whitespace?

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+    python-str    = "'" *(python-str-c / DQUOTE) "'"
+    python-str    =/ DQUOTE *(python-str-c / "'") DQUOTE
+    marker-var    = python-str / "python_version" / "python_full_version" /
+                    "os_name"" / "sys_platform" / "platform_release" /
+                    "platform_version" / "platform_machine" /
+                    "platform_python_implementation" / "implementation_name" /
+                    "implementation_version" / "platform_dist_name"
+                    "platform_dist_version" / "platform_dist_id"
+    marker-expr   = "(" marker ")" / (marker-var [marker-op marker-var])
+    marker        = marker-expr *( ("and" / "or") marker-expr)
```
njsmith: `marker-expr` and `marker` should allow whitespace

dependency-specification.rst
```diff
@@ -0,0 +1,376 @@
...
+    python-str-c  = (WSP / ALPHA / DIGIT / "(" / ")" / "." / "{" / "}" /
+                    "-" / "_" / "*"
+    python-str    = "'" *(python-str-c / DQUOTE) "'"
+    python-str    =/ DQUOTE *(python-str-c / "'") DQUOTE
+    marker-var    = python-str / "python_version" / "python_full_version" /
+                    "os_name"" / "sys_platform" / "platform_release" /
+                    "platform_version" / "platform_machine" /
+                    "platform_python_implementation" / "implementation_name" /
+                    "implementation_version" / "platform_dist_name"
+                    "platform_dist_version" / "platform_dist_id"
```
njsmith: What happens if we encounter an unrecognized variable name? Do we want to gracefully degrade somehow, e.g. by treating any marker expression that contains an unrecognized variable as evaluating to False? If so, then the grammar should probably be rewritten to something like
```
python-str = ...
marker-var = (ALPHA / "_") *(ALPHA / DIGIT / "_")
marker-val = python-str / marker-var
```
so that arbitrary python variables can at least be parsed, and then the valid ones specified below.

OTOH this may hide problems like @merwok's `os_name == posix` silently evaluating to False.

Probably it would be a good idea to only allow marker expressions of the form `<variable> <op> <literal>` (in exactly that order), but I don't know if that would break existing usage...

dependency-specification.rst
njsmith: It's unambiguous, at least, in that according to the grammar above `os_name==posix` is a parse error :-) An example wouldn't hurt.

rbtcollins: I believe its crystal clear from the first example, which shows a different variable without quotes, and the thing its compared to with quotes.

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: I'll a note. I'm being deliberately conservative - none of the vars contain escape-worthy characters today so there's no need for escapes.

dependency-specification.rst
rbtcollins: It guarantees that there will be at least one space after a url before a marker.

dependency-specification.rst
rbtcollins: I was missing * beside the CSP elements.

dependency-specification.rst
rbtcollins: yah, also addressed.

dependency-specification.rst
rbtcollins: gone, along with comments entirely.

dependency-specification.rst
rbtcollins: Addressed above.

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: SyntaxError - its what the behaviour is, and I'm not seeing to redefine markers, only to define the modern version + fix the one critical bug with 2.7.10.

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+Examples
+--------
+
+All features of the language shown with a name based lookup::
+
+    requests [security,tests] >= 2.8.1, == 2.8.* ; python_version < "2.7.10"
+
+A minimal URL based lookup::
+
+    pip @ https://github.com/pypa/pip/archive/1.3.1.zip#sha1=da9234ee9982d4bbb3c72346a6de940a148ea686
```
dstufft: An example that shows all of the features being used with URLs would be nice to include as well.

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+A distribution specification is written in ASCII text. We use ABNF [#abnf]_ to
+provide a precise grammar. The grammar covers just the specification itself.
+The expectation is that the specification will be embedded into a larger
+system which offers framing such as comments, multiple line support via
+continuations, or other such features.
+
+Versions may be specified according to the PEP-440 [#pep440]_ rules. (Note:
+URI is defined in std-66 [#std66]_::
+
+    version-cmp   = "<" / "<=" / "!=" / "==" / ">=" / ">"
```
dstufft: I think you're missing ``~=`` and ``===`` yea? We might also want to mention that any additional version specifiers added in future PEPs are implicitly included? I know I'd like to at least add ``^=`` at some point.

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+    version       = 1*( DIGIT / ALPHA / "-" / "_" / "." / "*" )
+    version-inner = version-cmp version *(',' version-cmp version)
+    versionspec   = ("(" version-inner ")") / version-inner
```
dstufft: Do we need to actually support the ``Foo (>=1.0)`` thing?

dependency-specification.rst
dstufft: I agree with an error though I don't think we should specify what that error is. That's a concern of the specific implementation how they represent that error. We should simply state that it is an error.

dstufft:
It'd be nice to include a bit of extra feature here. Particularly I'm thinking of:

* Add an implicit ``*`` extra which stands for "install all extras", e.g. ``requests[*]`.
  * This is an error today, so it shouldn't change any behaviors.
* Allow negating extras by prepending a ``-``. This is particular useful when combined with ``*``, you'd be able to do ``requests[*,-something]`` to get all of the extras except for the "something" extras.
  * Currently implements will implement this as converting it from ``-something`` to ``_something``. I don't think many (or any?) projects are using a leading underscore so I think that breakage will be minimal.

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+The marker language is a subset of Python itself, chosen for the ability to
+safely evaluate it without running arbitrary code that could become a security
+vulnerability. Markers were first standardised in PEP-345 [#pep345]_. This PEP
+fixes some issues that were observed in the described in PEP-426 [#pep426]_.
+
+Comparisons in marker expressions are typed by the comparison operator.  The
+<marker-op> operators that are not in <version-cmp> perform the same as they
+do for strings in Python. The <version-cmp> operators use the PEP-440
+[#pep440]_ version comparison rules if both sides are valid versions. If
+either side is not a valid version, then the comparsion falls back to the same
```
dstufft: I'm not sure if "is a valid version" is the right way to word this. ``1.*`` is not a valid version but it is a valid specifier. Perhaps this should take the same route that PEP 440 did and just say that they are always version comparisons and then let implementations handle the backwards compatibility concerns. In pip/setuptools that meant we fell back to the legacy behavior.

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+specification because none of the variables to which constants can be compared
+contain quote-requiring values.
+
+The variables in the marker grammar such as "os_name" resolve to values looked
+up in the Python runtime. With the exception of "extra" all values are defined
+on all Python versions today - it is an error in the implementation of markers
+if a value is not defined.
+
+Unknown variables must raise a SyntaxError rather than resulting in a
+comparison that evaluates to True or False.
```
dstufft: As I said above, I'd rather not declare a specific error. Just say that Unknown variables are an error and should not evaluate to either ``True`` or ``False``.

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+Most of this PEP is already widely deployed and thus offers no compatibiltiy
+concerns.
+
+There are however a few points where the PEP differs from the deployed base.
+
+Firstly, PEP-440 direct references haven't actually been deployed in the wild,
+but they were designed to be compatibly added, and there are no known
+obstacles to adding them to pip or other tools that consume the existing
+dependency metadata in distributions - particularly since they won't be
+permitted to be present in PyPI uploaded distributions anyway.
```
dstufft: The obstacle that is keeping them from being added is how do we prevent them from being uploaded to PyPI. Currently we can see the dependency information for a Wheel and can reject direct references in those (but we currently don't). However enforcing that a sdist doesn't use them is not currently possible and if @njsmith gets his wish where we never make it so that all possible dependencies for a sdist is declared up front in a static manner, then we might never be able to prevent them from being uploaded to PyPI inside of a sdist. Before we can implement them we need some answer to how we prevent them from being uploaded.

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+    marker-expr   = "(" *WSP marker *WSP ")"
+                  =/ (marker-var [*WSP marker-op *WSP marker-var])
+    marker        = *WSP marker-expr *( *WSP ("and" / "or") *WSP marker-expr)
+    name-marker   = ";" *WSP marker
+    url-marker    = ";" 1*WSP marker
+
+Optional components of a distribution may be specified using the extras
+field::
+
+    identifier    = 1*( DIGIT / ALPHA / "-" / "_" )
```
dstufft: PEP 426 (which wasn't ever accepted obviously) defined a name as having alpha numeric, ``-``, ``_``, and ``.``. It also stated that names must begin and end with a alpha numeric (IOW ``foo-bar`` is valid, but ``foobar-`` is not). It wasn't accepted but PyPI currently enforces that for names. I think that we should include that requirement here as well for extras too (outside of a leading ``-`` for negation).

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+
+    name_req      = name [*WSP extras] [*WSP versionspec] [*WSP name-marker]
+
+And a rule for direct reference specifications::
+
+    url_req       = name [*WSP extras] urlspec [*WSP url-marker]
+
+Leading to the unified rule that can specify a dependency::
+
+    specification = name_req / location_req
```
dstufft: This should probably support leading and trailing whitespace which is just ignored as well yea?

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+    marker        = *WSP marker-expr *( *WSP ("and" / "or") *WSP marker-expr)
+    name-marker   = ";" *WSP marker
+    url-marker    = ";" 1*WSP marker
+
+Optional components of a distribution may be specified using the extras
+field::
+
+    identifier    = 1*( DIGIT / ALPHA / "-" / "_" )
+    name          = identifier
+    extras        = "[" identifier *("," identifier) "]"
```
dstufft: Should it be valid to have empty extras? Like ``requests[]``?

dependency-specification.rst
```diff
@@ -0,0 +1,324 @@
...
+
+Environment markers allow making a specification only take effect in some
+environments::
+
+    marker-op     = version-cmp / "in" / "not in"
+    python-str-c  = WSP / ALPHA / DIGIT / "(" / ")" / "." / "{" / "}" /
+                    "-" / "_" / "*"
+    python-str    = "'" *(python-str-c / DQUOTE) "'"
+    python-str    =/ DQUOTE *(python-str-c / "'") DQUOTE
+    marker-var    = python-str / "python_version" / "python_full_version" /
```
dstufft: Am I reading this correctly that this allows something like ``"foobar" == "foobar"``? Under what situations is that useful?

dependency-specification.rst
njsmith: > It'd be nice to include a bit of extra features for extras. [`*` and `-`]

FWIW, adding these features now might rule out the option of ever compatibly migrating to my wacky extras-as-first-class-packages idea. I'm not saying we necessarily want to do that, but maybe we should wait on adding these features until we've had more time to work through the issues?

dstufft: I assume you meant to reply to a different comment ;) But answering your comment here:

Would it actually. Afaik it would just mandate that we need some method to query what the available extras are for a particular name. That doesn't seem like a particularly troublesome requirement.

njsmith: Ugh, yes, github's ui has defeated me somehow. Sorry.

Anyway: if extras were promoted to first class packages, then presumably the way you would query for what extras were available would be to ask your index server for a list of all the wheels it has whose names match a particular pattern. So `pip install django[*]` would be implemented the same way you'd implement `pip install django-*`, where they both mean "install all the packages that pypi knows of that have names matching this pattern". Maybe this makes sense in some cases for interactive use (and at least the trust issues would be mitigated in the django[*] case), but the technical issues mean that I really really don't think we'd want to support this inside a Requires-Dist. You couldn't even look at a local environment and check whether its packages have correctly satisfied dependencies satisfied without contacting a remote server. And if it'd only be for interactive use, then it doesn't need to go in a PEP anyway.

dstufft: Why could the hypothetical Django project not have to declare up front all of it's extras as part of it's own metadata?

dependency-specification.rst
dstufft: Also I think that this ws meant to be ``url_req`` not ``location_req``.

dependency-specification.rst
rbtcollins: Its just fugly cause of line length :/

dependency-specification.rst
rbtcollins: We had them in before, and removed them because of the interop problems they imply, if this is going to be used for solely new things, thats doable I think. But we'll not be solving the interop problem - packages using a new feature will blow up on older interpreters.

If we're adding new things to markers I think we should also add marker_version in the markers section to enable folk to opt in gracefully to future changes. Adding ^= would involve revving the marker version at the same time.

dependency-specification.rst
rbtcollins: pkg resources defines extras as A-Za-z0-9. only - other characters are collapsed to _

for other identifiers, its the same pattern but other characters are collapsed to -.

I think we should actually follow that rather than having logically same names appear different to humans.

dependency-specification.rst
rbtcollins: ok, sure.

dependency-specification.rst
rbtcollins: markers already permit that. The set of useful marker expressions is a subset of the marker expressions that can be written. ( ( ( python_version == python_version ) ) ) is also a valid expression.

dependency-specification.rst
rbtcollins: We can make it so; there's no semantic difference.

dependency-specification.rst
rbtcollins: done, done.

dependency-specification.rst
dstufft: Can we just not allow either of those? I doubt anyone is actually doing that right now (and once I'm done writing my implementation I can tell us how many people are actually doing it). It seems useful to turn something that is almost certainly a mistake into an error.

I guess it's used to also allow ``"2.7" == python_version`` but I don't honestly think that is very important either. We can test to see though. I think that unless there's some number of them doing that, we should just restrict it so the LHS can only be a variable name and the RHS can only be a string.

dependency-specification.rst
rbtcollins: Then we'd need to define operators to compare strings for e.g. "pos" < os_name. It may not make a lot of sense but its valid today, and leaving it unspecified is just letting the rot remain. If we want to make them prohibited, lets do that. Otherwise lets define it sensibly. - s/if both sides are valid versions/if the operators behaviours is defined given the value of both sides/

dstufft: Thinking about this more... we should probably split these based on whether or not the named item represents a version or not. If it doesn't represent a version than it uses string comparisons as implemented by Python. If it represents a version then it uses PEP 440 specifier syntax. This means that ``~=`` and such would only be valid for ones that actually represent versions.

dependency-specification.rst
rbtcollins: I don't think we can ever have 100% declarative dependencies. I think the fundamental thing should change - rather than preventing them being in PyPI, we should teach tools like pip to ignore them in things downloaded from PyPI.

dstufft: I remain unconvinced that we cannot, at the very least, define the possible dependencies up front. I have yet to hear a use case where people are expecting a package to have completely arbitrary dependencies. The ones I've heard seem to fall into two categories:

* We don't want to hard require certain things, but depending on features or what is available we might select different dependencies.
* We have a broad range for a dependency, but once we've built the binary then we need to restrict that to versions that will be ABI compatible with what we've built against.

Neither one of those are situations where the package authors can't enumerate the entire list of possible dependencies before hand, and then when the wheel gets built it can dynamically restrict the total items that it depends on from that total list of possibilities.

We don't really want to special case PyPI here, we've gone to some lengths to try and remove special casing PyPI other than the fact it's the default repository and we define things in terms of whether it's a public repository or a private repository.

That's neither here nor there though because this PEP can (and should) further build on the direct dependency syntax introduced in PEP 440. We just won't allow that particular part of the PEP in pip until we have a solution to being able to prevent people using them on PyPI (one way or another).

dependency-specification.rst
rbtcollins: Trying to restrict it has no utility; it will complicate implementations and make the grammar more complex.

dependency-specification.rst
rbtcollins: I don't see what you are saying as being different to what I wrote. Well defined PEP-440 behaviour for operators that PEP-440 defines when the two parameters are legitimate per PEP-440; fallback to something else otherwise. I defined 'something else' as 'Python native operator result or False', but we can pick something different.

dependency-specification.rst
rbtcollins: done.

dependency-specification.rst
rbtcollins: Sure; so nothing to see here, move along. That said, public index is something we can define and teach pip about.

dependency-specification.rst
dstufft: I think restricting it has utility, particularly for the version specifiers because they are designed to have the "Variable" portion be on the LHS and reverse that is going to give inconsistent or bad results, particularly for ``~=`` and ``==1.*`` and such.

I don't think it makes either the grammar or implementation more complex in a meaningful way. You just need to define something like:

```
marker-var-lhs "python_version" / "python_full_version" /
                       "os_name"" / "sys_platform" / "platform_release" /
                       "platform_version" / "platform_machine" /
                       "platform_python_implementation" / "implementation_name" /
                       "implementation_version" / "platform_dist_name"
                       "platform_dist_version" / "platform_dist_id"
marker-var-rhs python-str
marker-expr   = "(" *WSP marker *WSP ")"
                      =/ (marker-var-lhs [*WSP marker-op *WSP marker-var-rhs])
```

I don't think having to take one variable and split it into two and call them both once instead of one two is a meaningful amount of difference in complexity.

dependency-specification.rst
dstufft: I mean that it would be an error to do ``platform_machine ~= 'whatever'``. ``platform_machine`` is not a version.

njsmith: I think the idea (which I like) would be: our tiny language has two types: `string` and `version`. `a op b` always promotes `a` and `b` to a common type, and then selects an implementation of `op` based on this type. The promotion hierarchy is very simple: strings are upcastable to versions, but not vice-versa. Literals are always strings; the built-in variables are defined to be either strings (`os_name` etc.) or versions (`python_version` etc.) as appropriate.

End result is: `os_name < "pos"` always does string comparison, because both arguments have type `string`. `python_version >= "2.7.10"` always does version comparison because one of the arguments is a version. `python_version >= "pos"` is an error, because `"pos"` cannot be upcast to a version.

And once we're properly typed like this, we probably would want to also say that the operations defined on `string` are: `==`, `!=`, `in`, `not in`, and the operations defined on `version` are: `==`, `!=`, `<`, `<=`, `>`, `>=`. Trying to apply `in` to versions or `<` to strings is an error.

njsmith: ...or reading @dstufft's initial comment again, maybe we need 3 types, `string`, `version`, and `version_pattern`. Or maybe we don't need `version`, and just `string` + `version_pattern` are enough. Anyway, even if I don't 100% understand all the semantics of versions then I think the general approach makes sense :-)

dependency-specification.rst
dstufft: I'm not sure that ``marker_version`` makes sense. I guess that idea is that you could put that *first* in the markers and use short circuiting behavior to skip everything else, but that assumes that whatever changes are made are syntactically valid but semantically undefined or something. I'm trying to think of what cases that would actually help:

*  *A new variable*: Won't help because it'll fail trying to parse the markers if they contain a new variable.
* *A new specifier/operator*: Won't help because it'll fail trying to parse the markers if they contain the new operator.

The only case I can think of is cases where we *change* the meaning of one of the strings.. which seems largely unlikely since these strings are coming from Python not from the marker PEP/implementation.

dependency-specification.rst
dstufft: Throw in some ``...`` to shorten it to give an idea of what it looks like? IDK, might not be worth it.

dependency-specification.rst
njsmith: > We just won't allow that particular part of the PEP in pip until we have a solution to being able to prevent people using them on PyPI (one way or another).

I agree that pip shouldn't try to special-case PyPI or try to ignore metadata depending on its source. (@rbtcollins: The argument against the latter is that it messes up the resolution algorithm if the same metadata can specify different dependencies before its installed when it's read from PyPI, and after its installed when it's read from disk.) But I think the practical way forward here is for PyPI to disallow them in wheels, and not to worry about sdists? By the time you're installing from sdist then package authors can already force pip to install a particular version of some software. (They can also delete your home directory. Oh well.)

dependency-specification.rst
dstufft: Variable (LHS)                 | Supported Operators               | Comparison "Type"
------------------------------ | --------------------------------- | -----------------------
python_version                 | >, <, ==, !=, >=, <=, ===, ~=     | Version
implementation_version         | >, <, ==, !=, >=, <=, ===, ~=     | Version
platform_dist_version          | >, <, ==, !=, >=, <=, ===, ~=     | Version
python_full_version            | >, <, ==, !=, >=, <=, ===, ~=     | Version
platform_python_implementation | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
implementation_name            | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
platform_dist_name             | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
platform_release               | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
platform_version               | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
platform_machine               | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
platform_dist_id               | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
sys_platform                   | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String
os_name                        | >, <, ==, !=, >=, <=, in, not in  | Case Insensitive String

dependency-specification.rst
njsmith: New variables are okay if you move the variable lookup out of the actual grammar and into the interpreter -- right now in Python terms everything is a "keyword", but we could make them "variables" instead. ...But you're totally right about new operators.

Probably the way to handle this would be to add some extra piece of guard syntax that means "if you don't know about version X of the dependency specification syntax, then just ignore this whole requirement (= treat it like it's always satisfied)", like, I dunno,
```
(!!syntax>=1) requests ~= 2.3
```
with the idea being that because it's at the front and has a fixed form `r"\(!!syntax(<|>|<=|>=)([0-9]+)\)", you can check for it before you even parse the rest of the line. This is a terrible and line-noisey example of what the guard might look like, but you get the idea. It wouldn't help immediately because of the need to support old pips, but after a few years, once those are gone, we might be glad that we added it now.

Or we could just drop the idea altogether / defer it for a future PEP :-)

dependency-specification.rst
njsmith: Your example ABNF would allow `platform_version in "Oracle"` but not `"Oracle" in platform_version`. Could fix that by adding some rules so `in` and `not in` require a literal on the left and variable on the right, while the other operators are the other way around, but dunno if it's worth it.

dependency-specification.rst
njsmith: I don't really care either way, but isn't this a pointless TOOWTDI violation?

dependency-specification.rst
dstufft: I don't even really think that makes sense. "I depend on requests ~= 2.3 unless your version of pip is old" (alternatively, I need pywin32 on Windows... unless your version of pip is old)? I think the only sane thing to do is make it an error to attempt to install something which uses a marker "version" that you don't understand. Treating the variables as keywords means we get that for free except for when we change what the variables mean, which I think isn't very likely given we don't really define what these variables are.

dependency-specification.rst
dstufft: Yea sure, it's not super hard to allow that either. A more restricted grammar is easier to parse by naive methods.

dependency-specification.rst
njsmith: I guess the use case would be something like "I need some_module if (a) the machine doesn't support SSE3, or (b) pip isn't clever enough to know whether the machine supports SSE3":
```
Requires-Dist: (!!syntax>=2) some_module ; "SSE3" not in cpu_capabilities
Requires-Dist: (!!syntax<2) some_module
```

But I agree that this is pretty marginal, and am totally :+1: to the idea of reining in the scope of this PEP.

rbtcollins: Hmm, a marker_version can be done like so:

guarded_marker = [ "marker_version" *WSP ">=" *WSP version ] *WSP ( marker / *CHAR )

Markers are the end of the grammar - try to parse a valid marker in this specification, fallback to just eating up the end of the input.

That will deal with both new keywords, and new operators.

dstufft: And once you've decided you don't support the marker version what do you do?

dependency-specification.rst
rbtcollins: So yeah - language design is hard. The marker language already works. We don't need the grammar to only express useful things. We need it to be usable by users, which the current one will.

dependency-specification.rst
rbtcollins: @njsmith ?

dependency-specification.rst
njsmith: I am tempted by the "more restricted grammar" argument, but unfortunately don't find it super convincing given that the grammar also allows for arbitrarily parenthesized boolean expressions. Handling this spec is going to require Real Parser Technology regardless of whether we add restrictions on `variable op literal`.

dependency-specification.rst
njsmith: Allowing `requests[]` as an alias for `requests`.

dependency-specification.rst
rbtcollins: So, I still question this - its a large change from markers so far, based AFAICT on no actual data about problems.

Folk can write ```"2.7" in python_version``` sensibly today, and we're making that impossible by this sort of change. I'm very much against this speculative change: our goal is to describe what *is*, not to overhaul and do aspirational work.

In fact, I'm going to push back again on the addition of new things to markers. Its going to raise the work needed to roll stuff out, and its not related to thing we're trying to enable.

dependency-specification.rst
rbtcollins: @njsmith - so we track on disk whether something came from 'public index' or 'private'.

dependency-specification.rst
njsmith: But what started this whole subthread is that you (quite rightly) want to add version comparison semantics on top of what we have now, and so the question is how exactly to bolt this on :-). And in the process we're certainly going to break backcompat, in that e.g. `python_version >= "2.7.10"` might return something different before and after.

I can see an argument for defining `in` on version-typed objects if needed for backcompat, but I think the overall idea of being clear about types is better than ad-hoc fallback on error.

It would definitely help if @dstufft is able to generate a list of all the real-world-usages-seen-on-PyPI though :-).

dstufft: I think adding PEP 440 version comparisons needs to either be done holistically or not at all. I could see it being a bit more complicated though if we instead wanted to do something like:

For the "version" fields above, allow ``in`` and ``not in`` and using them does a case insensitive string comparison. Using any other operator does a version comparison/specifier as defined in PEP 440.

xavfernandez:
@dstufft adding [*] and [-exclude] syntax would require to describe how the union of extras should work.
If you have a package needing pkg[*,-extra] and an other one needing pkg[extra], do you install extra ?

dependency-specification.rst
njsmith: I guess if we're adding versioning it should be done for the whole thing, not just the environment marker? If we think we might want to add new comparison operators later, then that'll apply just as much to the versionspec part of the string as to the environment marker part of the string, so the version marker should go before both of them.

But probably the concrete action item for now is to add a note saying "we considered doing it this way ... but decided to defer that for later", so we can move forward without losing the design work.

njsmith:
Regarding wild card extras like `requests[*]`: it looks like part of that conversation has disappeared into some github-related black hole. My concern was that if we promote extras to first-class packages, then how do you resolve the `[*]`? @dstufft's suggestion was (IIUC) the metadata for the plain `requests` package would contain a field explicitly saying what `*` should be taken to expand to, so it'd be a kind of alias/indirection. That idea seems not-obviously-impossible to me, but I feel like this is getting tricky enough that we should defer it until we can have a proper conversation about it, instead of trying to piggyback on this PEP.

dstufft:
Seems fine to me.

dependency-specification.rst
rbtcollins: Donald and I had a quick call; the proposal is now:
 - ignore the lhs rhs thing
 - add all the operators in
 - my original suggested edit will cover the typing sufficiently.

dependency-specification.rst
rbtcollins: on extras - "-foo" meaning "not foo" needs defined semantics when combined cross specification.

If foo depends on bar and quux[*], and bar depends on quux[-baz], should quux[baz] be considered?

I'm going to defer negative extras for a future discussion because of that. I think including [*] can be done now though, since its strictly compatible with the unioning semantics.

dependency-specification.rst
rbtcollins: Discussed w/@dstufft - we agreed to add "." and also prevent identifiers starting or ending in non-alpha-numeric characters.

dependency-specification.rst
rbtcollins: How is that a TOOWTDI ?

rbtcollins:
I'm fine leaving the [*] out too - I'll do so until further reviews.

njsmith:
AFAIK the only motivation to adding special sugar for * inside pip is that
it lets you support the -exception syntax. The current ipython[all] hack
works fine as an equivalent to ipython[*]; it only breaks down if you want
ipython[*,-foo].

On Mon, Nov 9, 2015 at 4:33 PM, rbtcollins <notifications@github.com> wrote:

> I'm fine leaving the [*] out too - I'll do so until further reviews.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/pypa/interoperability-peps/pull/56#issuecomment-155245173>
> .
>



-- 
Nathaniel J. Smith -- http://vorpus.org

dependency-specification.rst
ncoghlan: I think Donald and Robert already resolved this, but a point worth noting: Python's *own* Grammar allows things that don't actually make any sense, and then we flag them as errors later in the code generation process.

A similar approach may be useful here - the formal grammar can be permissive, but as a quality of implementation issue, some tools may choose to flag apparent nonsense, like having constants on both sides of an operator ("2.7" == "2.6"), or variables on both sides (python_version == python_version).

dependency-specification.rst
ncoghlan: +1

Being able to define the restrictions on names formally in an accepted PEP is the main reason I suggested pulling dependency specifiers out as their own PEP, PEP 440 style.

dependency-specification.rst
dstufft: Yea, it was resolved by preferring compatibility with what's already there ("2.7" == "2.7" is already a valid marker) over trying to be overly protective from people doing silly, but generally harmless, things.

dependency-specification.rst
ncoghlan: It's not, it's the equivalent of allowing an empty base list in a class definition. Leaving out the brackets is still the obvious way to write it by hand, but allowing an empty extras list makes the degenerate case easy when the list of extras is produced programmatically with something like:

    "{dist}[{extras}]".format(dist=depname, extras=extras)

dependency-specification.rst
```diff
@@ -0,0 +1,326 @@
...
+Non line-breaking whitespace is mostly optional with no semantic meaning. The
+sole exception is detecting the end of a URL requirement.
+
+Names
+-----
+
+Python distribution names are currently defined in PEP-345 [#pep345]_. Names
+act as the primary identifier for distributions. They are present in all
+dependency specifications, and are sufficient to be a specification on their
+own.
```
ncoghlan: This is one of the aspects I'd like to see this spec tighten up over 345, by codifying the more restrictive rules defined in PEP 426: https://www.python.org/dev/peps/pep-0426/#name (I believe PyPI already enforces the more restrictive version)

dependency-specification.rst
ncoghlan: Do we want to define environment markers inline here, or give them their own PEP as in https://github.com/pypa/interoperability-peps/pull/29 ?

dependency-specification.rst
dstufft: Yes, PyPI mandates that all names must match this (case insensitive) regex: ``^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$``. That is enforced as a constraint in the database now. We also normalize the names using this (case insensitive) regex: ``lower(regexp_replace($1, '(\.|_)', '-', 'ig'))`` (PostgreSQL regex).

dependency-specification.rst
ncoghlan: I agree with @dstufft that the "support in this spec" and the "support in pip" questions can be separated for direct dependencies. Depending on software that has yet to be formally published is an interesting problem, as it has clear practical benefits *within* an organisation and during development, but poses challenges for dependencies *between* organisations and in released versions of software.

PEP 426 tries to formulate that distinction in terms of "public index server" vs "private index server", but I'm not entirely sure that's the right way of thinking about it. Fortunately, I think we can safely punt on it in relation to this particular spec - we know we want an answer to "How are direct dependencies written?", even though we're not entirely sure on "When we do want to accept direct dependencies on unpublished versions during installation?".

dependency-specification.rst
ncoghlan: Never mind, I've caught up on enough distutils-sig email to find the thread where you decided to inline it after discussing it with James.

rbtcollins: btw platform.dist is deprecated in 3.5, adding it to pip is madness. So I'm not putting those in as-is, but I will do platform.release() etc.

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+as comments, multiple line support via continuations, or other such features.
+
+The full grammar including annotations to build a useful parse tree is
+included at the end of the PEP.
+
+Versions may be specified according to the PEP-440 [#pep440]_ rules. (Note:
+URI is defined in std-66 [#std66]_::
+
+    version_cmp   = wsp* '<' | '<=' | '!=' | '==' | '>=' | '>' | '~=' | '==='
+    version       = wsp* ( letterOrDigit | '-' | '_' | '.' | '*' )+
```
dstufft: This can include a ``+`` too for local version identifiers as well as a ``!`` for epoch values.

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+    version_one   = version_cmp version
+    version_many  = version_one (wsp* ',' version_one)*
```
dstufft: Probably want a ``wsp*`` after the ``','`` as well to support ``>=1.0, <3`` which I don't think is currently supported.

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+    versionspec   = ( '(' version_many ')' ) | version_many
+    urlspec       = '@' wsp* <URI_reference>
+
+Environment markers allow making a specification only take effect in some
+environments::
+
+    marker_op     = version_cmp | 'in' | 'not in'
+    python_str_c  = (wsp | letter | digit | '(' | ')' | '.' | '{' | '}' |
```
dstufft: This also needs expanded since it can't contain all the values we need just for version numbers, but probably it should be expanded to handle more than just version numbers too yea? Right now you can't do something like ``platform_version == '#3 on degas'`` which is in the Python documentation as an example for that. This is also restricted to only ASCII characters, is that reasonable or is there a reasonable value for one of the variables we have below that doesn't fall inside ascii?

Either way, this probably does need to support all the printable ascii characters at a minimum and might need to handle backslash escapes for quotes too since a lot of these strings could be completely freeform.

dstufft: In particular, ``platform_version`` is the one that worries me the most about what may be there.

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+Variables whose value cannot be calculated on a given Python implementation
+should evaluate to ``0`` for versions, and an empty string for all other
+variables.
+
+The "extra" variable is special. It is used by wheels to signal which
+specifications apply to a given extra in the wheel ``METADATA`` file, but
+since the ``METADATA`` file is based on a draft version of PEP-426, there is
+no current specification for this. Regardless, outside of a context where this
+special handling is taking place, the "extra" variable should result in a
+SyntaxError like all other unknown variables.
```
dstufft: left over reference to an explicit error?

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+     - ``platform.python_version()``
+     - ``3.4.0``, ``3.5.0b1``
+   * - ``implementation_name``
+     - ``sys.implementation.name``
+     - ``cpython``
+   * - ``implementation_version``
+     - see definition below
+     - ``3.4.0``, ``3.5.0b1``
+   * - ``extra``
+     - A SyntaxError except when defined by the context interpreting the
```
dstufft: Another left over reference to a specific error?

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+       specification.
+     - ``test``
+
+The ``implementation_version`` marker variable is derived from
+``sys.implementation.version``::
```
dstufft: ``sys.implementation.version`` doesn't exist in Python 2.x, we should probably define what that looks like if ``sys.implementation.version`` doesn't exist.

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+
+Rationale
+=========
+
+In order to move forward with any new PEPs that depend on environment markers,
+we needed a specification that included them in their modern form. This PEP
+brings together all the currently unspecified components into a specified
+form.
+
+The requirement specifier EBNF is lifted from setuptools pkg_resources
```
dstufft: Left over reference to EBNF?

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+    marker_var    = env_var | python_str
+    marker_expr   = ('(' wsp* marker wsp* ')'
+                     | (marker_var wsp* marker_op wsp* marker_var))
+    marker        = wsp* marker_expr ( wsp* ('and' | 'or') wsp* marker_expr)*
+    quoted_marker = ';' wsp* marker
+
+Optional components of a distribution may be specified using the extras
+field::
+
+    identifier    = ( digit | letter | '-' | '_' | '.')+
```
dstufft: This was going to include mandating it start and end with an alpha num, so should probably be something more like:

```
identifier = letterOrDigit | letterOrDigit (letterOrDigit | '-' | '_' | '.')* letterOrDigit
```

That should allow single character identifiers that have a letter or digit, two character identifiers that are both letters and digits, and any identifier with 3+ characters that start and end with a letter or digit, and the middle ones are letters, digits, ``'-'``, ``'_'``, or ``'.'``.

dependency-specification.rst
```diff
@@ -0,0 +1,522 @@
...
+vulnerability. Markers were first standardised in PEP-345 [#pep345]_. This PEP
+fixes some issues that were observed in the described in PEP-426 [#pep426]_.
+
+Comparisons in marker expressions are typed by the comparison operator.  The
+<marker-op> operators that are not in <version_cmp> perform the same as they
+do for strings in Python. The <version_cmp> operators use the PEP-440
+[#pep440]_ version comparison rules when those are defined (that is when both
+sides have a valid version specifier). If there is no defined PEP-440
+behaviour and the operator exists in Python, then the operator falls back to
+the Python behaviour. Otherwise the result of the operator is False.
```
dstufft: Does this mean that ``python_version ~= 'dog'`` should evaluate as false rather than as an error?

dstufft: I'm thinking that maybe the fallback for not having a defined operation (either via PEP 440 or via Python) should be an error rather than silently being ``False``.

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: its supported via the leading wsp* on version_cmp

dependency-specification.rst
rbtcollins: we agreed on irc to drop platform_version and platform_release

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: prose added; didn't talk about the unique constraint, just the acceptability one.

dependency-specification.rst
rbtcollins: Sure

dependency-specification.rst
rbtcollins: done, though there is prose above to the same effect.

rbtcollins: (line 218)

dependency-specification.rst
```diff
@@ -0,0 +1,538 @@
...
+included at the end of the PEP.
+
+Versions may be specified according to the PEP-440 [#pep440]_ rules. (Note:
+URI is defined in std-66 [#std66]_::
+
+    version_cmp   = wsp* '<' | '<=' | '!=' | '==' | '>=' | '>' | '~=' | '==='
+    version       = wsp* ( letterOrDigit | '-' | '_' | '.' | '*' )+
+    version_one   = version_cmp version
+    version_many  = version_one (wsp* ',' version_one)*
+    versionspec   = ( '(' version_many ')' ) | version_many
```
njsmith: `= ( wsp* '(' version_many ')' ) | version_many` if you want to maintain the invariant that each rule absorbs leading whitespace.

njsmith: Or maybe `= ( wsp* '(' version_many wsp* ')' ) | version_many` since presumably `(>=1 )` should be just as legal as `( >=1)`.

dependency-specification.rst
```diff
@@ -0,0 +1,538 @@
...
+    urlspec       = '@' wsp* <URI_reference>
+
+Environment markers allow making a specification only take effect in some
+environments::
+
+    marker_op     = version_cmp | 'in' | 'not in'
```
njsmith: I guess if we're going to be super-nitpicky it should be `('not' wsp+ 'in')`

dependency-specification.rst
```diff
@@ -0,0 +1,538 @@
...
+    quoted_marker = ';' wsp* marker
+
+Optional components of a distribution may be specified using the extras
+field::
+
+    identifier    = ( letterOrDigit |
+                      letterOrDigit (letterOrDigit | '-' | '_' | '.')* letterOrDigit )
+    name          = identifier
+    extras_list   = identifier (wsp* ',' wsp* identifier)*
+    extras        = '[' wsp* extras_list? ']'
```
njsmith: add `wsp*` before the closing bracket

dependency-specification.rst
```diff
@@ -0,0 +1,538 @@
...
+
+    argparse;python_version<"2.7"
+
+A marker expression evalutes to either True or False. When it evaluates to
+False, the dependency specification should be ignored.
+
+The marker language is a subset of Python itself, chosen for the ability to
+safely evaluate it without running arbitrary code that could become a security
+vulnerability. Markers were first standardised in PEP-345 [#pep345]_. This PEP
+fixes some issues that were observed in the described in PEP-426 [#pep426]_.
```
njsmith: editing tangle: "observed in the described in"

dependency-specification.rst
```diff
@@ -0,0 +1,538 @@
...
+
+Comparisons in marker expressions are typed by the comparison operator.  The
+<marker-op> operators that are not in <version_cmp> perform the same as they
```
njsmith: I know this is super-nitpicky, but the mixture of hyphens and underscores in `marker-op` versus `version_cmp` is like nails on chalkboard for me. Maybe just pick a convention?

dependency-specification.rst
```diff
@@ -0,0 +1,538 @@
...
+behaviour and the operator exists in Python, then the operator falls back to
+the Python behaviour. Otherwise an error should be raised. e.g. the following
+will result in  errors::
+
+    "dog" ~= "fred"
+    python_version ~= "surprise"
+
+User supplied constants are always encoded as strings with either ``'`` or
+``"`` quote marks. Note that backslash escapes are not defined, but existing
+implementations do support them them. They are not included in this
```
njsmith: "them them"

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: they are meant to be all _, it was oversight.

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: done

dependency-specification.rst
rbtcollins: Covered by tweaking version_one

dependency-specification.rst
```diff
@@ -0,0 +1,543 @@
...
+:Post-History: XX
+
+
+Abstract
+========
+
+This PEP specifies the language used to describe dependencies for packages.
+It draws a border at the edge of describing a single dependency - the
+different sorts of dependencies and when they should be installed is a higher
+level problem. The intent is provide a building block for higher layer
```
xavfernandez: The intent is **to** provide ?

dependency-specification.rst
xavfernandez: Is it ? I don't see any `!` / `+`

dependency-specification.rst
```diff
@@ -0,0 +1,543 @@
...
+    version_one   = version_cmp version wsp*
+    version_many  = version_one (wsp* ',' version_one)*
+    versionspec   = ( '(' version_many ')' ) | version_many
+    urlspec       = '@' wsp* <URI_reference>
+
+Environment markers allow making a specification only take effect in some
+environments::
+
+    marker_op     = version_cmp | 'in' | 'not' wsp+ 'in'
+    python_str_c  = (wsp | letter | digit | '(' | ')' | '.' | '{' | '}' |
```
xavfernandez: Is there a difference between `letterOrDigit` and `letter | digit` ?

xavfernandez: and sorry if I missed it but how are `{` and `}` used in markers ? I've never seen those. Maybe an example using them would be nice ?

dependency-specification.rst
```diff
@@ -0,0 +1,543 @@
...
+                     | (marker_var wsp* marker_op wsp* marker_var))
+    marker        = wsp* marker_expr ( wsp* ('and' | 'or') wsp* marker_expr)*
+    quoted_marker = ';' wsp* marker
+
+Optional components of a distribution may be specified using the extras
+field::
+
+    identifier    = letterOrDigit (
+                    letterOrDigit |
+                    (( letterOrDigit | '-' | '_' | '.')* letterOrDigit ) )*
```
xavfernandez: Are identifiers starting with a digit allowed ?

dependency-specification.rst
```diff
@@ -0,0 +1,543 @@
...
+
+Python distribution names are currently defined in PEP-345 [#pep345]_. Names
+act as the primary identifier for distributions. They are present in all
+dependency specifications, and are sufficient to be a specification on their
+own. However, PyPI places strict restrictions on names - they must match a
+case insensitive regex or they won't be accepted. Accordingly in this PEP we
+limit the acceptable values for identifiers to that regex. A full redefinition
+of name may take place in a future metadata PEP::
+
+    ^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$
```
xavfernandez: This seems odd, maybe some parentheses are missing ?

dependency-specification.rst
```diff
@@ -0,0 +1,543 @@
...
+
+An extra is an optional part of a distribution. Distributions can specify as
+many extras as they wish, and each extra results in the declaration of
+additional dependencies of the distribution **when** the extra is used in a
+dependency specification. For instance::
+
+    requests[security]
+
+Extras union in the dependencies they define with the dependencies of the
+distribution they are attached to. The example above would result in requests
```
xavfernandez: `Extras union in the dependencies they define with the dependencies of the distribution they are attached to.`
This sentence seems odd, maybe the **in** shouldn't be there ?

dependency-specification.rst
```diff
@@ -0,0 +1,543 @@
...
+A marker expression evalutes to either True or False. When it evaluates to
+False, the dependency specification should be ignored.
+
+The marker language is a subset of Python itself, chosen for the ability to
+safely evaluate it without running arbitrary code that could become a security
+vulnerability. Markers were first standardised in PEP-345 [#pep345]_. This PEP
+fixes some issues that were observed in the design described in PEP-426
+[#pep426]_.
+
+Comparisons in marker expressions are typed by the comparison operator.  The
```
xavfernandez: Extra space here: `operator.  The`

dependency-specification.rst
```diff
@@ -0,0 +1,543 @@
...
+
+    if hasattr(sys, 'implementation'):
+        implementation_version = format_full_version(sys.implementation.version)
+    else:
+        implementation_version = "0"
+
+Backwards Compatibility
+=======================
+
+Most of this PEP is already widely deployed and thus offers no compatibiltiy
```
xavfernandez: compatibility

dependency-specification.rst
rbtcollins: skew between the complete grammar and this fragment; fixed.

dependency-specification.rst
```diff
@@ -0,0 +1,556 @@
...
+    version_many  = version_one (wsp* ',' version_one)*
+    versionspec   = ( '(' version_many ')' ) | version_many
+    urlspec       = '@' wsp* <URI_reference>
+
+Environment markers allow making a specification only take effect in some
+environments::
+
+    marker_op     = version_cmp | (wsp* 'in') | (wsp* 'not' wsp+ 'in')
+    python_str_c  = (wsp | letter | digit | '(' | ')' | '.' | '{' | '}' |
+                     '-' | '_' | '*')
```
xavfernandez: ok, so `:` and `#` (at least) are also missing.
cf your example: `#1 SMP Fri Apr 25 13:07:35 EDT 2014`

rbtcollins: So I was trying to stay away from embedding quoting in here - I'd rather add it later if we can. We can add more punctuation other than \\ very easily though.

dependency-specification.rst
ncoghlan: It's easy to miss the pipe operator in the middle there. The LHS is special casing single alphanumeric character distribution names (which are weird, but technically permitted), the RHS is the more typical case of "2+ characters, must start and end with an alphanumeric character, may contain periods, hyphens and underscores".

rbtcollins: I want to use the exact regex PyPI uses; should I add some prose or a human description?

dstufft: FTR PyPI uses it with ``re.IGNORECASE`` flag.

rbtcollins: noted in prose

dependency-specification.rst
```diff
@@ -0,0 +1,561 @@
...
+
+Versions may be specified according to the PEP-440 [#pep440]_ rules. (Note:
+URI is defined in std-66 [#std66]_::
+
+    version_cmp   = wsp* '<' | '<=' | '!=' | '==' | '>=' | '>' | '~=' | '==='
+    version       = wsp* ( letterOrDigit | '-' | '_' | '.' | '*' | '+' | '!' )+
+    version_one   = version_cmp version wsp*
+    version_many  = version_one (wsp* ',' version_one)*
+    versionspec   = ( '(' version_many ')' ) | version_many
+    urlspec       = '@' wsp* <URI_reference>
```
qwcode: everything else in the grammar is atomic and well defined.   "URI_reference" is not.   It's a big loaded concept.

it seems to me you need another line almost to define it.

```
URI_reference = <A valid PEP440 "Direct Reference" URL>
```

rbtcollins: Huh? URI_reference is a single rule in std-66 - its very well defined.

qwcode: I thought you were trying to mean PEP440 Direct references using `<URI_reference>`

it wasn't clear what you're referring to

why the carets? and the spelling from std66 is "URI-reference"  (https://tools.ietf.org/html/std66#section-4.1)

rbtcollins: What carets?

qwcode: >  why the carets

I mean, why the `<` and `>` in `<URI_reference>`

rbtcollins: Thats 'capture all the things the rule matched' in OMeta. Having it up here is a typo, since its really intended for the combined grammar, but its not harmful, and doesn't change the things that are accepted by the grammar.

PR #60, opened 2015-12-08
PEP 440: Remove PEP 426 references

s-t-e-v-e-n-k:
PEP 426 is in Draft status, and will not be accepted. PEP 508 is Accepted, and is the way forward, so replace references of 426 with 508 inside PEP 440.

[Review on Reviewable](https://reviewable.io/reviews/pypa/interoperability-peps/60)


pep-0440-versioning.rst
```diff
@@ -1084,7 +1103,7 @@ uploaded distributions. Direct references are intended as a tool for
 software integrators rather than publishers.
 
 Depending on the use case, some appropriate targets for a direct URL
-reference may be a valid ``source_url`` entry (see PEP 426), an sdist, or
+reference may be a valid ``source_url`` entry (see PEP 508), an sdist, or
```
rbtcollins: I think this is actually just a gone concept: direct references can only be sdists or wheels now.

PR #62, opened 2016-01-05
PEP 508: Switch to platform_python_implementation

s-t-e-v-e-n-k:
python_implementation as a marker environment variable in PEP 508 should be
spelt platform_python_implementation, correct it.

[Review on Reviewable](https://reviewable.io/reviews/pypa/interoperability-peps/62)


rbtcollins:
looks good to me

PR #63, opened 2016-02-17
Another build system abstraction PEP

njsmith:
[Review on Reviewable](https://reviewable.io/reviews/pypa/interoperability-peps/63)

### DIFF
--- a/unpublished/build-system-abstraction-njs.rst
+++ b/unpublished/build-system-abstraction-njs.rst
@@ -1,0 +1,708 @@
+PEP: ??
+Title: A build-system independent format for source trees
+Version: $Revision$
+Last-Modified: $Date$
+Author: Nathaniel J. Smith <njs@pobox.com>
+Status: Draft
+Type: Standards-Track
+Content-Type: text/x-rst
+Created: 30-Sep-2015
+Post-History: 1 Oct 2015, 25 Oct 2015
+Discussions-To: <distutils-sig@python.org>
+
+==========
+ Abstract
+==========
+
+While ``distutils`` / ``setuptools`` have taken us a long way, they
+suffer from three serious problems: (a) they're missing important
+features like usable build-time dependency declaration,
+autoconfiguration, and even basic ergonomic niceties like `DRY
+<https://en.wikipedia.org/wiki/Don%27t_repeat_yourself>`_-compliant
+version number management, and (b) extending them is difficult, so
+while there do exist various solutions to the above problems, they're
+often quirky, fragile, and expensive to maintain, and yet (c) it's
+very difficult to use anything else, because distutils/setuptools
+provide the standard interface for installing packages expected by
+both users and installation tools like ``pip``.
+
+Previous efforts (e.g. distutils2 or setuptools itself) have attempted
+to solve problems (a) and/or (b). This proposal aims to solve (c).
+
+The goal of this PEP is get distutils-sig out of the business of being
+a gatekeeper for Python build systems. If you want to use distutils,
+great; if you want to use something else, then that should be easy to
+do using standardized methods. The difficulty of interfacing with
+distutils means that there aren't many such systems right now, but to
+give a sense of what we're thinking about see `flit
+<https://github.com/takluyver/flit>`_ or `bento
+<https://cournape.github.io/Bento/>`_. Fortunately, wheels have now
+solved many of the hard problems here -- e.g. it's no longer necessary
+that a build system also know about every possible installation
+configuration -- so pretty much all we really need from a build system
+is that it have some way to spit out standard-compliant wheels and
+sdists.
+
+We therefore propose a new, relatively minimal interface for
+installation tools like ``pip`` to interact with package source trees
+and source distributions.
+
+
+=======================
+ Terminology and goals
+=======================
+
+A *source tree* is something like a VCS checkout. We need a standard
+interface for installing from this format, to support usages like
+``pip install some-directory/``.
+
+A *source distribution* is a static snapshot representing a particular
+release of some source code, like ``lxml-3.4.4.zip``. Source
+distributions serve many purposes: they form an archival record of
+releases, they provide a stupid-simple de facto standard for tools
+that want to ingest and process large corpora of code, possibly
+written in many languages (e.g. code search), they act as the input to
+downstream packaging systems like Debian/Fedora/Conda/..., and so
+forth. In the Python ecosystem they additionally have a particularly
+important role to play, because packaging tools like ``pip`` are able
+to use source distributions to fulfill binary dependencies, e.g. if
+there is a distribution ``foo.whl`` which declares a dependency on
+``bar``, then we need to support the case where ``pip install bar`` or
+``pip install foo`` automatically locates the sdist for ``bar``,
+downloads it, builds it, and installs the resulting package.
+
+Source distributions are also known as *sdists* for short.
+
+A *build frontend* is a tool that users might run that takes arbitrary
+source trees or source distributions and builds wheels from them. The
+actual building is done by each source tree's *build backend*. In a
+command like ``pip wheel some-directory/``, pip is acting as a build
+frontend.
+
+An *integration frontend* is a tool that users might run that takes a
+set of package requirements (e.g. a requirements.txt file) and
+attempts to update a working environment to satisfy those
+requirements. This may require locating, building, and installing a
+combination of wheels and sdists. In a command like ``pip install
+lxml==2.4.0``, pip is acting as an integration frontend.
+
+
+==============
+ Source trees
+==============
+
+We retroactively declare the legacy source tree format involving
+``setup.py`` to be "version 0". We don't try to specify it further;
+its de facto specification is encoded in the source code and
+documentation of ``distutils``, ``setuptools``, ``pip``, and other
+tools.
+
+A "version 1" (or greater) source tree is any directory which contains
+a file named ``pypackage.json``. (If a tree contains both
+``pypackage.json`` and ``setup.py`` then it is a version 1+ source
+tree and the ``setup.py`` is ignored; this allows packages to include
+a ``setup.py`` for compatibility with old build frontends, while using
+the new system with new build frontends.)
+
+This file has the following schema. Extra keys are ignored.
+
+schema
+    The version of the schema. This PEP defines version "1".  Defaults to "1"
+    when absent. All tools reading the file MUST error on an unrecognised
+    schema version.
+
+bootstrap_requires
+    Optional list of PEP 508 dependency specifications that the
+    build frontend must ensure are available before invoking the build
+    backend. For instance, if using flit, then the requirements might
+    be::
+
+        "bootstrap_requires": ["flit"]
+
+build_backend
+    A mandatory string naming a Python object that will be used to
+    perform the build (see below for details). This is formatted
+    following the same ``module:object`` syntax as a ``setuptools``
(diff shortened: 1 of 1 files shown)


rbtcollins:
So there's a lot in common twixt our proposals and some differences.

- .dist-info directory on disk. This is ugly, and has no benefits I can see. pip has no need for any data that isn't already in METADATA for the planning and scheduling of the install.

- calling interface convention. You list three advantages, but I dispute that two of them are advantages  - they are at most neutral:

1. code reuse: we can put any interface code we want to reuse in a Python library and use that in flit and other build systems, leading to the same amount of code. I am also skeptical of the claim that there is enough common code in the 'child cmdline interface' to matter - since I've now written one, and 99% of it is marshalling command line arguments.

2. While Python is indeed a great language, the claim that CLI interfaces are going to be intrinsically worse is unsupported IMO. Python interfaces that call out to external commands (as I expect every thunk to do)  are going to have to deal with all the complexities of incremental output, dealing with stdout and stderr from the things they called, concurrency etc. We're not avoiding that, we're just failing to deal with it. Python code can just write to pip's output for instance, which is even more intrusive - as pip has no control over that!

Your third point has some merit - and in fact my original PEP had a negotiation step which @ Donald's request I removed :/. That said, compatible changes we can add by try-and-fallback as long as we define a clear fallback mechanism - which I just realise isn't defined (e.g. a standard exit code like 77).

- passing information between metadata and wheel is unneeded IMO - it just introduces room for pip to introduce bugs in builds.

njsmith:
> pip has no need for any data that isn't already in METADATA for the planning and scheduling of the install.

Not right now. But pip might change, and the metadata might change... "What does pip 8.0 need" isn't a great principle for defining this interface I think :-/. And in practice the metadata in a wheel is the whole directory, not just the one file. Alternatively I considered trying to actually define the metadata that we think is needed early, but this quickly becomes nasty

rbtcollins:
pip has about 8 years of data leading up to the data it needs: what pip 8 uses is actually a very good guide to what metadata is needed. And since we're going to be evolving wheel's metadata format too since its currently defacto-defined, we can take backwards compat into account as we evolve it. Better than yet another YAGNI PEP IMO.

njsmith:
Doh sorry, my pocket seems to have posted the first unedited part of my reply.

rbtcollins:
Pocket dialing, 2016 edition :)

unpublished/build-system-abstraction-njs.rst
```diff
@@ -0,0 +1,708 @@
...
+documentation of ``distutils``, ``setuptools``, ``pip``, and other
+tools.
+
+A "version 1" (or greater) source tree is any directory which contains
+a file named ``pypackage.json``. (If a tree contains both
+``pypackage.json`` and ``setup.py`` then it is a version 1+ source
+tree and the ``setup.py`` is ignored; this allows packages to include
+a ``setup.py`` for compatibility with old build frontends, while using
+the new system with new build frontends.)
+
```
pfmoore: Unlike Robert's proposal, this one explicitly *does* get into the business of defining a sdist format (via defining the "source tree" format). And it does so in a way that ignores the possibility of static (= introspectable without needing to involve the build system) metadata. And what does it gain us to be any more formal than Robert's PEP? ("The build interface is defined in terms of a source tree that contains a single config file and "other stuff" as needed by the build system"). We don't provide any benefits to tools that *don't* need to do builds (consider twine, which in uploading a sdist may benefit from being able to extract the metadata and add it to the PyPI project metadata - but which doesn't want to go near the build system).

I still think I'd like to see a formal sdist format at some point, simply so that we have a reference for how tools can introspect any file they find on PyPI, but the discussion on Robert's PEP has convinced me that build system abstraction proposals are not the right place to get into that.

unpublished/build-system-abstraction-njs.rst
```diff
@@ -0,0 +1,708 @@
...
+The build backend object is expected to have attributes which provide
+some or all of the following hooks. The common ``config_settings``
+argument is described after the individual hooks::
+
+  def get_build_requires(config_settings):
+      ...
+
+This hook MUST return an additional list of strings containing PEP 508
+dependency specifications, above and beyond those specified in the
+``pypackage.json`` file. Example::
```
pfmoore: Is it acceptable for this to return values that depend on the environment the build tool is being run on?

unpublished/build-system-abstraction-njs.rst
```diff
@@ -0,0 +1,708 @@
...
+Must create a ``.dist-info`` directory containing wheel metadata
+inside the specified ``metadata_directory`` (i.e., creates a directory
+like ``{metadata_directory}/{package}-{version}.dist-info/``. This
+directory MUST be a valid ``.dist-info`` directory as defined in the
+wheel specification, except that it need not contain ``RECORD`` or
+signatures. The hook MAY also create other files inside this
+directory, and a build frontend MUST ignore such files; the intention
+here is that in cases where the metadata depends on build-time
+decisions, the build backend may need to record these decisions in
+some convenient format for re-use by the actual wheel-building step.
```
pfmoore: Again, is it allowable for the metadata to depend on the build environment?

unpublished/build-system-abstraction-njs.rst
```diff
@@ -0,0 +1,708 @@
...
+undefined, but basically comes down to: a file named
+``{NAME}-{VERSION}.{EXT}``, which unpacks into a buildable source tree
+called ``{NAME}-{VERSION}/``. Traditionally these have always
+contained "version 0" source trees; we now allow them to also contain
+version 1+ source trees.
+
+Integration frontends require that an sdist named
+``{NAME}-{VERSION}.{EXT}`` will generate a wheel named
+``{NAME}-{VERSION}-{COMPAT-INFO}.whl``.
+
```
pfmoore: I do like the fact that you formally record the requirement on the metadata (name, version) embedded in the sdist filename here. I don't actually think that's recorded in a (finalised) spec anywhere else.

unpublished/build-system-abstraction-njs.rst
```diff
@@ -0,0 +1,708 @@
...
+and thus expose a totally uniform interface to the rest of ``pip``,
+with no extra subprocess calls, no duplicated builds, etc. But
+obviously this is the kind of code that you only want to write as part
+of a private, within-project interface.
+
+(And, of course, making the ``metadata`` command optional is one piece
+of lowering the barrier to entry, as discussed above.)
+
+
```
pfmoore: One downside of a Python-based interface is that it makes it harder to write build systems that aren't implemented in Python. I'm not saying this is necessarily a good thing to do, but in the spirit of your comment above about quick hacks developing into significant tools, I could imagine someone wanting to write a primitive build tool around something like make and a few shell scripts.

Of course you can't specify "I need make and bash" as a PEP 508 style requirement, so in practice the fact that you have to be able to indicate the build system using ```bootstrap_requires``` precludes such options anyway.

unpublished/build-system-abstraction-njs.rst
```diff
@@ -0,0 +1,708 @@
...
+possible for build backends to define both methods, for compatibility
+with both old and new build frontends.
+
+Furthermore, our mechanism should also fulfill two more goals: (a) If
+new versions of e.g. ``pip`` and ``flit`` are both updated to support
+the new interface, then this should be sufficient for it to be used;
+in particular, it should *not* be necessary for every project that
+*uses* ``flit`` to update its individual ``pypackage.json`` file. (b)
+We do not want to have to spawn extra processes just to perform this
+negotation, because process spawns can easily become a bottleneck when
```
xavfernandez: negotiation

unpublished/build-system-abstraction-njs.rst
njsmith: Yes, it might (and in some cases surely will) depend on OS, python version, the `config_settings`, etc. That's why it's a function rather than a static list :-)

unpublished/build-system-abstraction-njs.rst
njsmith: Yeah, ditto. E.g. numpy will almost certainly support being built in different configurations for no-BLAS (= using slow internal fallback code), using external system-provided BLAS (which currently gets the same metadata as the previous, but might get different metadata in the future if the externel dependencies PEP goes forward), and getting BLAS from a wheel (which would lead to an explicit requirement for that wheel in the metadata).

unpublished/build-system-abstraction-njs.rst
njsmith: thanks, fixed

unpublished/build-system-abstraction-njs.rst
njsmith: > One downside of a Python-based interface is that it makes it harder to write build systems that aren't implemented in Python.

It's true, and probably this should get some discussion in the PEP. My feeling though is that this is basically a good thing. If you really want to use shell then it's not hard to thunk from Python back to the shell -- it's a roadblock, but a minor one. And I think this is a case where a minor roadblock is actually a small net positive, if anything. The big problem with build systems is bootstrapping -- e.g. the entire reason for `autoconf`'s horrible architecture is that they start from the assumption that the build machine has *nothing* available except for an ancient `/bin/sh` and cruddy versions of related tools. (And of course this creates even worse problems these days because Windows is an important target platform and doesn't even have that.) Knowing that we have python available is a *huge* win for build system authors, and there's literally nothing else that they *can* assume is available. So really python is the right language to be working in unless you have a really good reason otherwise :-). So I'm not bothered if there's a small technical nudge pushing people in that direction.

And of course we anticipate that python will overwhelmingly be the most popular language for writing these systems anyway (cf. distutils, setuptools, flit, bento), so I'd rather optimize for that case anyway.

unpublished/build-system-abstraction-njs.rst
pfmoore: So if it's a pure Python project, and I build a wheel on my Windows box, it's acceptable for the function to return Windows-specific requirements?

Ah, wait. This is *build* requirements. So these requirements won't be needed at runtime. My apologies, I see your point now. I was confused. I'll leave it to you to decide if my confusion was reasonable, and it's worth clarifying the docs at all, or if I was just being dumb. I favour the latter explanation :-)

unpublished/build-system-abstraction-njs.rst
pfmoore: ... and in this case a different wheel *name* presumably (because having two copies of ```foo-1.0-cp35-none-win_amd64.whl``` with different requirements metadata is wrong).

You can of course do this with ```setup.py``` at the moment, so we're not making the situation any worse here. But at *some* point we need to thrash out how to document that metadata can only differ if the name, version or compatibility tag data differs. We can defer that to a later date, though, I guess.

unpublished/build-system-abstraction-njs.rst
pfmoore: Agreed.

dstufft:
Much like Robert's PEP, I'm going to merge this and then give @njsmith commit so he can update it inside of the repository itself and assign it a number.

dstufft:
This is now PEP 517.

unpublished/build-system-abstraction-njs.rst
njsmith: Yeah, unfortunately right now the name etc will be the same because there simply is no alternative -- e.g. these packages all satisfy a downstream requirement for `numpy`. This is one reason I'm keen to see the "reified extras" idea implemented, because among other things it would give us a framework for distinguishing these build variants.

PR #64, opened 2016-07-07
Update PEP 517 to use pyproject.toml from PEP 518

takluyver:
Ping @njsmith @rbtcollins.

* Switch `pypackage.json` to `pyproject.toml` and change description accordingly.
* Make `build_backend` key optional, so that any projects already using `pyproject.toml` following PEP 518 are not invalid.

I'm happy to make similar changes to PEP 516 if that's desired, but I'm not 100% sure if this is the right place, so I thought I'd start with one PEP.

---
This change is [Reviewable](https://reviewable.io/reviews/pypa/interoperability-peps/64)

### DIFF
--- a/pep-0517-build-system-abstraction.rst
+++ b/pep-0517-build-system-abstraction.rst
@@ -99,63 +99,53 @@ specification is encoded in the source code and documentation of
 ``distutils``, ``setuptools``, ``pip``, and other tools. We'll refer
 to it as the ``setup.py``\-style.
 
-Here we define a new ``pypackage.json``\-style source tree. This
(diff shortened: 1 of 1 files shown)


takluyver:
Should I be sending these changes somewhere else now? I'm not that familiar with how the PEP system works.

ncoghlan:
The PEPs repo itself is now on GitHub, and at least @rbtcollins should be able to accept PRs there directly: https://github.com/python/peps

takluyver:
Thanks Nick, I'll send a PR there.

